Issue #11, opened 2024-06-18
Rebranding


zuiderkwast:
Rename the files, functions, structs, comments.

* [ ] standalone (non-cluster) code
* [ ] cluster code

PR #18, opened 2024-06-19
Rebranding code in libvalkey

michael-grunder:
I left `Redis` in-tact in things like the `CHANGELOG.md` and all the copyright headers but think I got everything else.

Part of #11


bjosv:
In the PR for the cluster code Viktor had a fair [comment](https://github.com/valkey-io/libvalkey/pull/15#discussion_r1646348942) about renaming files multiple times. Maybe we should move the renamed files to src/ directly here too?

michael-grunder:
To `libvalkey/src` in this PR or to the root?

libvalkey/async.c
```diff
@@ -359,129 +359,129 @@ static void __redisRunDisconnectCallback(redisAsyncContext *ac, int status)
...
     char *stype;
     sds sname = NULL;
 
     /* Match reply with the expected format of a pushed message.
      * The type and number of elements (3 to 4) are specified at:
-     * https://redis.io/topics/pubsub#format-of-pushed-messages */
-    if ((reply->type == REDIS_REPLY_ARRAY && !(c->flags & REDIS_SUPPORTS_PUSH) && reply->elements >= 3) ||
-        reply->type == REDIS_REPLY_PUSH) {
-        assert(reply->element[0]->type == REDIS_REPLY_STRING);
+     * https://valkey.io/topics/pubsub#format-of-pushed-messages */
```
bjosv, line 481: The currently working link is 
```suggestion
     * https://valkey.io/docs/topics/pubsub/#format-of-pushed-messages */
```
but this might be a moving target..

libvalkey/test.sh
```diff
@@ -1,43 +1,43 @@
...
 ENABLE_DEBUG_CMD=
 SSL_TEST_ARGS=
 SKIPS_ARG=${SKIPS_ARG:-}
-REDIS_DOCKER=${REDIS_DOCKER:-}
+VALKEY_DOCKER=${VALKEY_DOCKER:-}
 
-# We need to enable the DEBUG command for redis-server >= 7.0.0
-REDIS_MAJOR_VERSION="$(redis-server --version|awk -F'[^0-9]+' '{ print $2 }')"
-if [ "$REDIS_MAJOR_VERSION" -gt "6" ]; then
+# We need to enable the DEBUG command for valkey-server >= 7.0.0
```
bjosv, line 13: There is no Valkey < 7.2.5, so we should always add `enable-debug-command local`

bjosv: not sure, but maybe we want to be able to test with redis here, then we could use `$VALKEY_SERVER --version...` to get the version instead

bjosv:
> To `libvalkey/src` in this PR or to the root?

To `<repo root>/src/`, then we would have them in the right place from start. Then I'll move the cluster source there aswell so we have them mixed.

libvalkey/test.sh
bjosv: That could be a separate PR, just mentioning..

libvalkey/adapters/glib.h
```diff
@@ -1,63 +1,63 @@
-#ifndef __HIREDIS_GLIB_H__
-#define __HIREDIS_GLIB_H__
+#ifndef VALKEY_GLIB_H
+#define VALKEY_GLIB_H__
```
bjosv, line 2:
```suggestion
#define VALKEY_GLIB_H
```

bjosv: Another on here, sorry. The comment at the end of file seem right.

zuiderkwast (approved):
LGTM

Reviewed in a scrolling speed of 1 page/second. :sunglasses:

Let's merge this and fix stuff afterwards? I believe that allows more people to work in parallel.

Some things to consider in the future follow-ups:

* Move fuzzing to `/tests/`
* Move examples to `/examples/` or are they actually tests?
* What include files to we like the users to see?
  * `<valkey/valkey.h>` sync API, e.g. valkeyCommand()
  * `<valkey/async.h>` async API, e.g. valkeyAsyncCommand()
  * `<valkey/cluster.h>` Cluster sync API, e.g. valkeyClusterCommand()
  * `<valkey/cluster-async.h>` e.g. valkeyClusterAsyncCommand()? (or is it -AsyncCluster-?, i forgot)
* Move exported API include files to /include/? (Doesn't affect where gets installed.)
  * Adapters .h files to /include/adapters/?
* Rename SSL to TLS?
* Or is it better to change as little as possible, so users can predictably just change "redis" to "valkey" and then everything else will work out of the box?

michael-grunder:
Sounds good.  Feel free to merge.

PR #21, opened 2024-06-21
Fix ISO C naming violations of function names and header guards

### LINKED ISSUE #19: ISO C naming of functions and header guards
- Rename functions starting with `_` or `__`
   Example: `_valkeyClusterConnect`
- Rename headerguards
  Remove `__` or `_` and match to filename.
  Example:  src/valkeycluster.h: `__HIRCLUSTER_H` to `VALKEYCLUSTER_H`

bjosv:
* Update header guards after move of files
* Fix ISO-C naming violation by renaming internal functions:
  * ..SetError()
  *  _valkeyClusterConnect()
  * __valkeyClusterAppendCommand()
  * __valkeyClusterGetReplyFromNode()
* Remove internal func __valkeyClusterGetReply()
* Remove internal func _valkeyClusterConnect2()

Closes #19


libvalkeycluster/tests/test_utils.h
```diff
@@ -1,5 +1,5 @@
-#ifndef __TEST_UTILS_H__
```
zuiderkwast, line 1: Why add VALKEY to each of these? Do you think there's a risk of macro name collission?

bjosv: Probably no collision in src/, I mostly followed how its done in the non-cluster headers.
```
src/async_private.h: #ifndef VALKEY_ASYNC_PRIVATE_H
src/fmacros.h:       #ifndef VALKEY_FMACRO_H
```
One (only?) benefit I can think of is that its already correct if we need to make a header public.

zuiderkwast (approved):

PR #24, opened 2024-06-23
Build complete library

michael-grunder:
This is an initial restructuring of the project to use an idiomatic linux shared object structure.

```bash
── examples
│   ├── CMakeLists.txt
│   └── Makefile
│   ├── examples1.c
│   ├── examples2.c
├── include
│   └── valkey
│       ├── adapters
│       │   ├── adapter1.h
│       │   ├── adapter2.h
│       ├── valkey.h
│       ├── other-public-headers.h
├── Makefile
├── src
│   └── internal.h
│   └── sources.c
└── tests
    ├── test.c
    ├── test1.c
    └── test.sh
```


Makefile
```diff
@@ -146,108 +166,42 @@ ifeq ($(uname_S),Darwin)
...
-$(SSL_OBJ): ssl.c valkey.h read.h sds.h alloc.h async.h win32.h async_private.h
-#################### SSL building rules end ####################
+$(SSL_STLIBNAME): $(SSL_OBJS)
+	$(STLIB_MAKE_CMD) $(SSL_STLIBNAME) $(SSL_OBJS)
 
-# Binaries:
-valkey-example-libevent: examples/example-libevent.c adapters/libevent.h $(STLIBNAME)
-	$(CC) -o examples/$@ $(REAL_CFLAGS) -I. $< -levent $(STLIBNAME) $(REAL_LDFLAGS)
+$(OBJ_DIR)/%.o: $(SRC_DIR)/%.c | $(OBJ_DIR)
+	$(CC) -std=c99 $(REAL_CFLAGS) -I$(INCLUDE_DIR) -MMD -MP -c $< -o $@
```
bjosv, line 184: Hurray for generated dependencies :)

Makefile
```diff
@@ -146,108 +166,42 @@ ifeq ($(uname_S),Darwin)
...
 
-valkey-example-libevent-ssl: examples/example-libevent-ssl.c adapters/libevent.h $(STLIBNAME) $(SSL_STLIBNAME)
-	$(CC) -o examples/$@ $(REAL_CFLAGS) -I. $< -levent $(STLIBNAME) $(SSL_STLIBNAME) $(REAL_LDFLAGS) $(SSL_LDFLAGS)
+$(OBJ_DIR)/%.o: $(TEST_DIR)/%.c | $(OBJ_DIR)
```
bjosv, line 186: I see a Makefile in tests/, what is this for?

Makefile
```diff
@@ -311,20 +253,22 @@ $(SSL_PKGCONFNAME): valkey_ssl.h
 	@echo Libs.private: -lssl -lcrypto >> $@
 
 install: $(DYLIBNAME) $(STLIBNAME) $(PKGCONFNAME) $(SSL_INSTALL)
```
bjosv, line 255: I wonder if there is a good way to create helper functions in a Makefile for these install/pkgconfig targets?
We will have for of them when cluster is incorporated, but maybe being explicit is the best option..

bjosv (commented):
Some early comment/questions, but I think it's very good in general.
I'll look at it a bit more.

Makefile
michael-grunder: It's going.away.  It was initially how I structured things.

Makefile
michael-grunder: Not sure, will look into it.

Makefile
bjosv, line 1: The file move linkage to the old libvalkey/Makefile existed before, not sure if this is a Github issue..

michael-grunder:
Non-cluster tests are now passing on Ubuntu, CentOS and RockyLinux.

The macOS failures are actually also happening in hiredis so we'll have to investigate what's going on there.

I can probably also get a windows VM set up to try and get that building again, but I know nothing about windows.

I think we can merge this and then start in on combining the unit tests?

PR #25, opened 2024-06-24
Remove parsing of unused cluster slot information

bjosv:
Currently there is no usage of the parsed `importing` and `migrating` information,
which is only provided by `CLUSTER NODES`.
Let's remove this superfluous code since we rely on `ASK` redirections,
which will be emitted when a slot is in `migrating` state.


zuiderkwast (approved):
Great.

We can still correctly ignore this information in the response, I hope. Do we have any test covering that?

bjosv:
> We can still correctly ignore this information in the response, I hope. Do we have any test covering that?

There are no tests covering that as I know, but I added #33 to improve that.
When inspecting the code we can still ignore the info though.

PR #26, opened 2024-06-25
Move remaining tests and build/run all legacy tests using CMake

### LINKED ISSUE #12: Move tests to `tests/`
Put all tests under `tests/` and fix how to build them and run them for cluster and non-cluster.

bjosv:
This PR moves the remaining tests, i.e. the cluster tests, to the new project structure
and enables building and running all tests using CMake.
```
      mkdir build && cd build
      cmake -DENABLE_SSL=ON -DENABLE_IPV6_TESTS=ON ..
      make
      make test
```

The cluster tests are a bit more complex to build and run than the non-cluster,
so currently the repo provided Makefile will only build the non-cluster tests.
This can probably be improved in the near future.

The legacy CMake options `ENABLE_SSL_TESTS` and `ENABLE_ASYNC_TESTS `are removed.
This is now instead controlled by the build option `ENABLE_SSL` and if `libevent` is available.

Closes #12


bjosv:
Ouch, I couldn't add the binary `test` from `test.c` as a target in CMake since its already used by `ctest`.

I had to set the outputname like `set_target_properties(libvalkey-test PROPERTIES OUTPUT_NAME "test")`
..which works with Make but not with ninja, as it seems.
I didn't test ninja locally but of course it now fails in CI..

michael-grunder (approved):
Looks good

bjosv:
Naming suggestions for the binary are welcomed, naming is the hard part :)

michael-grunder:
> Naming suggestions for the binary are welcomed, naming is the hard part :)

client_test works for me.  It is mostly what those tests are doing.

PR #27, opened 2024-06-26
Touchup makefile

michael-grunder:
* Remove dead assignment.
* Pass optimization level to the link-time step as well.


bjosv (approved):
Good findings

PR #29, opened 2024-06-26
Fix CI issues

bjosv:
- Fixes to build on FreeBSD in CI:
    - Header `alloca.h` is not available on FreeBSD.
    - Only add `libevent` includes if found.
    - Don't build tests in CI job on FreeBSD when using CMake.
    - Skip test/examples using `glib` if `pkg-config` is not installed.
    - Remove unused test label UT/CT.

- Run CI workflow "C/C++ CI"

- Get redis_version if valkey_version missing in INFO.
    Fixes issues in CI job "DB compatibility testing"


michael-grunder (approved):
Awesome!  Looking forward to more green tests.

PR #34, opened 2024-06-27
Pin Github Action versions in CI 

bjosv:
Pin the Github Actions we use in CI to a release hash according to secure software development best practices,
recommended by the Open Source Security Foundation (OpenSSF).

When developing a CI workflow, it's common to version-pin dependencies (i.e. actions/checkout@v4).
However, version tags are mutable, so a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead. Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.

See details:
https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -173,7 +173,7 @@ jobs:
         run: |
           choco install -y ninja memurai-developer
 
-      - uses: ilammy/msvc-dev-cmd@v1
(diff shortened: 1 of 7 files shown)


michael-grunder (approved):
Those crafty hackers :smile:

PR #38, opened 2024-06-29
Rename and remove deprecated functions in the cluster API

bjosv:
Rename API functions to get a standalone context from a cluster node, which are needed to support PUB/SUB.
```
ctx_get_by_node  -> valkeyClusterGetValkeyContext
actx_get_by_node -> valkeyClusterGetValkeyAsyncContext
```
- Remove internal `parse_cluster_*` functions from the cluster API. Should not be needed by users.
- Remove deprecated functions in the cluster API.

### DIFF
--- a/include/valkey/valkeycluster.h
+++ b/include/valkey/valkeycluster.h
@@ -172,10 +172,6 @@ void valkeyClusterFree(valkeyClusterContext *cc);
 /* Configuration options */
 int valkeyClusterSetOptionAddNode(valkeyClusterContext *cc, const char *addr);
 int valkeyClusterSetOptionAddNodes(valkeyClusterContext *cc, const char *addrs);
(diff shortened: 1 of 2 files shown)


zuiderkwast (commented):
Looks good!

Do you want to update the docs markdown file in the same PR?

We should remember this change for the migration guide. We could create a label for the PR for that.

bjosv:
The current docs don't mention `xx_get_by_node` as these APIs shouldn't be needed by the common user.
We should add it to the migration guide though, which I'm working on. The labels might be a good start and the release-drafter needs labels too. I'll look at it.

zuiderkwast:
How is this a breaking change? We haven't released anything yet.

zuiderkwast (approved):

PR #40, opened 2024-06-30
Minor Makefile cleanup

michael-grunder:
* We can specify SHELL = /bin/sh to avoid having to do `$(shell sh -c '<cmd>')` and instead just doing `$(shell <cmd)`.
* Use makefile built-in `word` to parse library version info.
* Minor cleanup of OS specific conditional blocks.
* Remove dead `CXX` assignment (it was required for the qt example).
* re-enable `-pedantic` for FreeBSD

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@
 # Copyright (C) 2010-2011 Pieter Noordhuis <pcnoordhuis at gmail dot com>
 # This file is released under the BSD license, see the COPYING file
 
+SHELL := /bin/sh
+
 SRC_DIR = src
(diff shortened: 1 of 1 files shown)


bjosv (approved):
LGTM!
If we keep SunOS support its might be good that we look into a CI for it (for a rainy day).
Maybe https://github.com/vmactions/solaris-vm can be used.

michael-grunder:
> If we keep SunOS support its might be good that we look into a CI for it (for a rainy day)

Good idea. I know people use Hiredis in some exotic environments (like AIX).

PR #41, opened 2024-07-01
Cleanup of unused and superfluous code in `src/vkutil.*`

bjosv:
- Remove unused code in `src/vkutil.*`
- Remove macro `NOT_REACHED()`.
- Remove function `vk_valid_port()`.
- Remove own assert macro in cluster code.
  We have a mixed usage of own and standard assert. Lets use standard from <assert.h> only.


zuiderkwast:
Failing jobs?

bjosv:
> Failing jobs?

Yes, seem to be some issue with windows that needs to be corrected. CentOS fails due to some install server

bjosv:
Yes, we should replace CentoOS 7 and run clang-tidy on everything, I has never passed in libvalkey after the merge.
Lets add issue on those two things

bjosv:
Added #42 and #43

PR #49, opened 2024-07-15
Fix license for new headers.

### LINKED ISSUE #48: Don't use Redis as project name in new files license text
We should not create new files with a license mentioning Redis as the name of the project. Let's change this in a follow-up.

A trademark-neutral version of the license mentions "the copyright holder" here instead of Redis.

_Originally posted by @zuiderkwast in https://github.com/valkey-io/libvalkey/pull/47#discussion_r1677478572_

michael-grunder:
IANAL so not sure if @pizhenwei
 would need to author the commit but this fixes #48

### DIFF
--- a/src/vkutil.h
+++ b/src/vkutil.h
@@ -13,9 +13,9 @@
  *   * Redistributions in binary form must reproduce the above copyright
  *     notice, this list of conditions and the following disclaimer in the
  *     documentation and/or other materials provided with the distribution.
- *   * Neither the name of Redis nor the names of its contributors may be used
- *     to endorse or promote products derived from this software without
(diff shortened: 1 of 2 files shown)


src/vkutil.h
```diff
@@ -13,9 +13,9 @@
  *   * Redistributions in binary form must reproduce the above copyright
  *     notice, this list of conditions and the following disclaimer in the
  *     documentation and/or other materials provided with the distribution.
- *   * Neither the name of Redis nor the names of its contributors may be used
```
pizhenwei, line 16: The commit message describes https://github.com/valkey-io/libvalkey/issues/48 only, is more description needed?

PR #53, opened 2024-07-17
Improve formatting

mkmkme:
I've tweaked `.clang-format` file to my liking and added a script that formats all the files.

Please don't hesitate to add any suggestions, comments, improvements.


michael-grunder:
LGTM

bjosv (commented):
Good work, I think the output from the formatter is ok enough to use.
There are a lot of more configs in valkey clang-format that we could try, but by keeping the defaults we make it simple. A formatter is what it is, we'll never get fully happy.

scripts/format-files.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
```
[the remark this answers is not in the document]
mkmkme: Yeah pinning clang-format version is something that has to be done one way or another. It's notorious for breaking things between different versions.

include/valkey/adapters/ae.h
```diff
@@ -30,11 +30,11 @@
 
 #ifndef VALKEY_AE_H
 #define VALKEY_AE_H
-#include <sys/types.h>
-#include <ae.h>
-#include "../valkey.h"
 #include "../async.h"
+#include "../valkey.h"
 #include "../valkeycluster.h"
```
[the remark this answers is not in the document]
mkmkme: Yeah we might add a new include category or change the include paths for `adapters` directory.

include/valkey/valkey.h
```diff
@@ -154,17 +158,21 @@ struct valkeySsl;
...
-                                          * on an explicit call to disconnect()
-                                          * or free() */
-#define VALKEY_OPT_NO_PUSH_AUTOFREE 0x08  /* Don't automatically intercept and
-                                          * free RESP3 PUSH replies. */
-#define VALKEY_OPT_NOAUTOFREEREPLIES 0x10 /* Don't automatically free replies. */
-#define VALKEY_OPT_PREFER_IPV4 0x20       /* Prefer IPv4 in DNS lookups. */
-#define VALKEY_OPT_PREFER_IPV6 0x40       /* Prefer IPv6 in DNS lookups. */
-#define VALKEY_OPT_PREFER_IP_UNSPEC (VALKEY_OPT_PREFER_IPV4 | VALKEY_OPT_PREFER_IPV6)
+
+/* Don't automatically free the async object on a connection failure, or
```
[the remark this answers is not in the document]
mkmkme: To be honest, I didn't manage to properly format it, for some reason even the formatter itself gets quite intrusive, like it breaks the string
```c
#define VALKEY_OPT_PREFER_IPV6 0x40 /* Prefer IPv6 in DNS lookups. */
```
to something like
```c
#define VALKEY_OPT_PREFER_IPV6                                  \
    0x40  /* Prefer IPv6 in DNS lookups. */
```

which IMHO is very ugly and I couldn't find a setting to prevent this from happening

michael-grunder:
This will need to be rebased against HEAD but are there any other blocking changes? 

Only downside is I can't spend half of my time on code formatting, but that's OK

mkmkme:
> This will need to be rebased against HEAD but are there any other blocking changes?

Rebased. I wonder if we want to enforce the formatting in the new PRs as a part of this PR or we can introduce it later.

mkmkme:
Hold on. I forgot to format `include` directory. Will do it now.

include/valkey/valkey.h
```diff
@@ -158,17 +162,18 @@ struct valkeySsl;
...
-                                          * on an explicit call to disconnect()
-                                          * or free() */
-#define VALKEY_OPT_NO_PUSH_AUTOFREE 0x08  /* Don't automatically intercept and
-                                          * free RESP3 PUSH replies. */
-#define VALKEY_OPT_NOAUTOFREEREPLIES 0x10 /* Don't automatically free replies. */
-#define VALKEY_OPT_PREFER_IPV4 0x20       /* Prefer IPv4 in DNS lookups. */
-#define VALKEY_OPT_PREFER_IPV6 0x40       /* Prefer IPv6 in DNS lookups. */
-#define VALKEY_OPT_PREFER_IP_UNSPEC (VALKEY_OPT_PREFER_IPV4 | VALKEY_OPT_PREFER_IPV6)
+#define VALKEY_OPT_NOAUTOFREE                                                  \
+    0x04 /* Don't automatically free the async object on a connection failure,
```
mkmkme, line 166: This change I personally dislike. We can discuss the ways to fix it but on my previous iteration I did it by moving the comment above this line.

michael-grunder: Moving above the line seems like a good solution.

.clang-format
```diff
@@ -0,0 +1,22 @@
+BasedOnStyle: LLVM
```
pizhenwei, line 1: What about explicit `ColumnLimit`?

pizhenwei:
Hi @mkmkme 
Should we update code check rule in Github workflow?

src/async.c
```diff
@@ -63,59 +65,58 @@ static unsigned int callbackHash(const void *key) {
...
-    callbackHash,
-    NULL,
-    callbackValDup,
-    callbackKeyCompare,
-    callbackKeyDestructor,
-    callbackValDestructor
+    callbackHash,          NULL,
+    callbackValDup,        callbackKeyCompare,
+    callbackKeyDestructor, callbackValDestructor,
 };
```
zuiderkwast, line 105: Yeah, this is a bit ugly, but iut's in very few places so I think it's OK.

We can maybe avoid it if make some manual changes to the code, like if we spell out the fields explicitly,

```C
    .hashFunction = callbackHash,
    .keyDup = NULL,
    ...
```

zuiderkwast (approved):
I scrolled through most of the re-formatted code. I'm fine with this style.

src/async.c
pizhenwei: If we use clang-format to check code format of the manual changes, it will report error. Or other folks run clang-format codes, it will be re-formatted. So I guess keeping the few places would be acceptable.

bjosv (commented):
Format is fine with me, just update so its builds and fix https://github.com/valkey-io/libvalkey/pull/53#discussion_r1725384594 
then I believe this should be good to go.
Kicking of checks in CI might require some extra configs needed, and can be done in a separate PR.

mkmkme:
@pizhenwei ,

> Hi @mkmkme Should we update code check rule in Github workflow?

Yup, thanks, I'll enable it

mkmkme:
The failure looks odd. My `clang-format` version is 18.1.6. The one in CI seems to be 18.1.3. I'll try to get 18.1.3 locally and see if it complains.

mkmkme:
So... The formatting indeed differs just a bit between 18.1.3 and 18.1.6...  That's bad news.

Good news is that you can install pretty much any clang-format version via `pip` on your machine: `pip install clang-format==18.1.3`

The github action uses the version from Ubuntu. I'd assume that this version is the most popular due to Ubuntu being the most popular distro.
Hence, I'd propose the following:
1. I'll reformat it with 18.1.3 (it will change like 3 lines of code in `tests` in comparison to the current format)
2. We'll explicitly state that the code should be formatted with 18.1.3

This is not ideal, but this is always the case with `clang-format` :/

examples/async-qt.cpp
```diff
@@ -1,16 +1,17 @@
 #include <iostream>
 using namespace std;
 
+#include "example-qt.h"
```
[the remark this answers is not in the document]
bjosv: Fixed in separate PR

include/valkey/valkey.h
```diff
@@ -42,14 +43,15 @@
...
 #include "alloc.h" /* for allocation wrappers */
+#include "sds.h"   /* for sds */
+
+#include <stdint.h> /* uintXX_t, etc */
 
 #define LIBVALKEY_MAJOR 1
 #define LIBVALKEY_MINOR 2
 #define LIBVALKEY_PATCH 0
-#define LIBVALKEY_SONAME 1.2.1-dev
+#define LIBVALKEY_SONAME 1.2.1 - dev
```
bjosv, line 54: This is a problem, only seems to trigger failures in the windows builds for some reason

mkmkme: Yup, this one is only used on Windows. I'll try putting quotes there just in case we want to use it anywhere else. If that doesn't fix Windows builds, I'll mark this line as ignored by formatter.

mkmkme:
Okay, all builds are finally green :) Let me know if anything else needs improving. If not, I'll squash some of the commits

.clang-format
```diff
@@ -0,0 +1,23 @@
...
+AlignTrailingComments: Always
+IncludeBlocks: Regroup
+IncludeCategories:
+  # fmacros.h must always be first since it defines macros that affect the
+  # other includes
+  - Regex: '^"(fmacros|win32).h"'
+    Priority: -1
+  - Regex: '^"'
+    Priority: 1
+  - Regex: '^<(valkey/adapters)/'
```
bjosv, line 16: Should be enough?
```suggestion
  - Regex: '^<valkey/adapters/'
```

.clang-format
```diff
@@ -0,0 +1,23 @@
...
+    Priority: 3
+  - Regex: '^<(valkey)/'
```
bjosv, line 18:
```suggestion
  - Regex: '^<valkey/'
```

.clang-format
```diff
@@ -0,0 +1,23 @@
+BasedOnStyle: LLVM
+ColumnLimit: 0
+IndentWidth: 4
+BreakBeforeTernaryOperators: false
+ReflowComments: false
+AllowAllArgumentsOnNextLine: true
+AlignTrailingComments: Always
+IncludeBlocks: Regroup
+IncludeCategories:
+  # fmacros.h must always be first since it defines macros that affect the
```
bjosv, line 10: Since we added win32.h here now;
something like "Headers that must be first.."

bjosv (approved):

PR #57, opened 2024-07-30
windows: fix size of `ssize_t`

mkmkme:
Previously, `long long` was used unconditionally to define `ssize_t` on Windows. This meant it was always a 64-bit value. However, typically the size type is a 32-bit value on 32-bit machines and 64-bit value on 64-bit machines.

This didn't cause any issues until it was used on 32-bit Windows build of `libvalkey-py`. CPython defines the `ssize_t` type as well as `libvalkey`, but it defines the size to be 32 bit on 32 bit machines which caused a compilation error.

This commit changes the typedef to be `SSIZE_T` on Windows. Hence, it will be aligned with everything else.

### DIFF
--- a/include/valkey/sds.h
+++ b/include/valkey/sds.h
@@ -35,8 +35,13 @@
 
 #define SDS_MAX_PREALLOC (1024*1024)
 #ifdef _MSC_VER
-typedef long long ssize_t;
-#define SSIZE_MAX (LLONG_MAX >> 1)
+#include <basetsd.h>
+typedef SSIZE_T ssize_t;
+#ifdef _WIN64
+#define SSIZE_MAX (_I64_MAX >> 1)
+#else
+#define SSIZE_MAX (_I32_MAX >> 1)
+#endif
 #ifndef __clang__
 #define __attribute__(x)
 #endif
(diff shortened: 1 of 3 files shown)


mkmkme:
Code format checker as well as macOS and Centos build failures look unrelated

include/valkey/sds.h
```diff
@@ -35,7 +35,8 @@
 
 #define SDS_MAX_PREALLOC (1024*1024)
 #ifdef _MSC_VER
-typedef long long ssize_t;
+#include <basetsd.h>
+typedef SSIZE_T ssize_t;
 #define SSIZE_MAX (LLONG_MAX >> 1)
```
bjosv, line 40: Should `SSIZE_MAX` also be changed in some similar way?
LLONG_MAX seems to big on your 32bit..or maybe it will "just work" ?

mkmkme: Oh yeah thanks, good catch. I'll fix it.

mkmkme: Fixed

bjosv (approved):

mkmkme:
I don't think I have merge rights. If you think it's good to go, please merge so I can update it in libvalkey-py

PR #59, opened 2024-08-01
Correcting source- and headerfile filtering in Makefile

bjosv:
Fix issue after the RDMA feature delivery; filter out ssl source files again.
   Issue found in [CI](https://github.com/valkey-io/libvalkey/actions/runs/10194328609/job/28200692136#step:4:22).

Filter out valkeycluster_ssl.h and only install it when built with USE_SSL.
 
Additionally update CI to use SSL when verifying the installation.


michael-grunder (approved):
LGTM

michael-grunder:
Apologies, I totally missed the regression because we don't have all-green tests yet.

bjosv:
No worries, I could see that and started to fix CI. Hopefully it will be better after #60 is in

PR #61, opened 2024-08-04
Initial readme.

michael-grunder:
Just the non-cluster parts of the code so far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,68 @@
+# Libvalkey
+
+Libvalkey is the official C client for the [Valkey](https://valkey.io) database. It also supports any server that uses the `RESP` protocol (version 2 or 3). This project supports both standalone and cluster modes.
+
+## Table of Contents
+
+- [Features](#features)
+- [Supported platforms](#supported-platforms)
+- [Building](#building)
+  - [Building with make](#building-with-make)
+  - [Building with CMake](#building-with-cmake)
+- [Contributing](#contributing)
+- Using the library
+  - [Standalone mode](docs/standalone.md)
+  - [Cluster mode](docs/cluster.md)
+
+## Features
+
+- Commands are executed in a generic way, with printf-like invocation.
+- Supports both `RESP2` and `RESP3` protocol versions.
+- Supports both synchronous and asynchronous operation.
+- Optional support for `SSL` and `RDMA` connections.
+- Asynchronous API with several event libraries to choose from.
+- Supports both standalone and cluster mode operation.
+- Can be compiled with either `make` or `CMake`.
+
+## Supported platforms
+
+This library supports and is tested against `Linux`, `FreeBSD`, `macOS`, and `Windows`. It should build and run on various proprietary Unixes although we can't run CI for those platforms. If you encounter any issues, please open an issue.
+
+## Building
+
+Libvalkey is written in C targeting C99. Unfortunately we have no plans on supporting C89 or earlier. The project does use a few widely supported compiler extensions, specifically for the bundled `sds` string library, although we have plans to remove this from the library.
+
+We support plain GNU make and CMake. Following is information on how to build the library.
+
+### Building with make
+
+```bash
+# Build and install the default library
+sudo make install
+
+# With all options
+sudo USE_SSL=1 USE_RDMA=1 make install
+
+# If your openssl is in a non-default location
+sudo USE_SSL=1 OPENSSL_PREFIX=/path/to/openssl make install
+```
+
+### Building with CMake
+
+See [CMakeLists.txt](CMakeLists.txt) for all available options.
+
+```bash
+# Build and install
+mkdir build && cd build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
+sudo make install
+
+# Build with TLS and RDMA support
+mkdir build && cd build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_SSL=1 -DENABLE_RDMA=1 ..
+sudo make install
+```
+
+## Contributing
+
+Contributions from the community are welcome. Feel free to open a PR with a bugfix or new feature. Note that if you would like to merge a large feature it's probably a good idea to open an issue first rather than spending a lot of time on a PR that may not be accepted.
--- a/docs/standalone.md
+++ b/docs/standalone.md
@@ -1,0 +1,352 @@
+# Standalone API documentation
+
+This document describes using `libvalkey` in standalone (non-cluster) mode, including an overview of the synchronous and asynchronous APIs. It is not intended as a complete reference. For that it's always best to refer to the source code.
+
+## Table of Contents
+
+- [Synchronous API](#synchronous-api)
+  - [Connecting](#connecting)
+  - [Connection options](#connection-options)
+  - [Executing commands](#executing-commands)
+  - [Using replies](#using-replies)
+  - [Reply types](#reply-types)
+  - [Disconnecting / cleanup](#disconnecting--cleanup)
+  - [Pipelining](#pipelining)
+  - [Errors](#errors)
+  - [Thread safety](#thread-safety)
+  - [Reader configuration](#reader-configuration)
+    - [Input buffer size](#maximum-input-buffer-size)
+    - [Maximum array elements](#maximum-array-elements)
+    - [RESP3 Push Replies](#resp3-push-replies)
+    - [Allocator injection](#allocator-injection)
+- [Asynchronous API](#asynchronous-api)
+  - [Connecting](#connecting-1)
+  - [Executing commands](#executing-commands-1)
+
+## Synchronous API
+
+The synchronous API has a pretty small surface area, with only a few commands to use. In general they are very similar to the way printf works, except they construct `RESP` commands.
+
+### Connecting
+
+There are several convenience functions to connect in various ways (e.g. host and port, Unix socket, etc). See [include/valkey.h](include/valkey.h) for more details.
+
+```c
+valkeyContext *valkeyConnect(const char *host, int port);
+valkeyContext *valkeyConnectUnix(const char *path);
+
+// There is also a convenience struct to specify various options.
+valkeyContext *valkeyConnectWithOptions(valkeyOptions *opt);
+```
+
+When connecting to a server, libvalkey will return `NULL` in the event that we can't allocate the context, and set the `err` member if we can connect but there are issues. So when connecting it's simple to handle error states.
+
+```c
+valkeyContext *ctx = valkeyConnect("localhost", 6379);
+if (ctx == NULL || ctx->err) {
+    fprintf(stderr, "Error connecting: %s\n", ctx ? ctx->errstr : "OOM");
+}
+```
+
+### Connection options
+
+There are a variety of options you can specify when connecting to the server, which are delivered via the `valkeyOptions` helper struct. This includes information to connect to the server as well as other flags.
+
+```c
+valkeyOptions opt = {0};
+
+// You can set primary connection info
+if (tcp)  {
+    VALKEY_OPTIONS_SET_TCP(&opt, "localhost", 6379);
+} else {
+    VALKEY_OPTIONS_SET_UNIX(&opt, "/tmp/valkey.sock");
+}
+
+// You may attach any arbitrary data to the context
+VALKEY_OPTIONS_SET_PRIVDATA(&opt, my_data);
+```
+
+There are also several flags you can specify when using the `valkeyOptions` helper struct.
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -0,0 +1,375 @@
...
+
+```bash
+# Build and install
+mkdir build && cd build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
+sudo make install
+
+# Build with TLS and RDMA support
+mkdir build && cd build
+cmakd -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_SSL=1 -DENABLE_RDMA=1 ..
```
pizhenwei, line 74: Typo `cmakd`.

README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Libvalkey is written in C targeting C99. Unfortunately we have no plans on supporting C89 or earlier. The project does use a few widely supported compiler extensions, specifically for the bundled `sds` string library, although we have plans to remove this from the library.
+
+We support plain GNU make and CMake. Following is information on how to build the library.
+
+### Building with make
+
+```bash
+# Build and install the default library
+make & sudo make install
```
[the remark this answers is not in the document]
bjosv: or alternatively just `sudo make install` as used in below examples.

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
+# standalone API documentation
```
bjosv, line 1:
```suggestion
# Standalone API documentation
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+  - [Connection options](#connection-options)
+  - [Executing commands](#executing-commands)
+  - [Using replies](#using-replies)
+  - [Reply types](#reply-types)
+  - [Disconnecting / cleanup](#disconnecting--cleanup)
+  - [Pipelining](#pipelining)
+  - [Errors](#errors)
+  - [Thread safety](#thread-safety)
+  - [Reader configuration](#reader-configuration)
+    - [Input buffer size](#maxinput-buffer-size)
```
bjosv, line 18: Not sure, but I believe the link should be
```suggestion
    - [Input buffer size](#maximum-input-buffer-size)
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+There are a variety of options you can specify when connecting to the server, which are delivered via the `valkeyOptions` helper struct. This includes information to connect to the server as well as other flags.
+
+```c
+valkeyOptions opt = {0};
+
+// You can set primary connection info
+if (tcp)  {
+    VALKEY_OPTIONS_SET_TCP(&opt, "localhost", 6379);
+} else {
+    VALKEY_OPTIONS_SET_UNIX(&opt, "/tmp/redis.sock");
```
bjosv, line 62:
```suggestion
    VALKEY_OPTIONS_SET_UNIX(&opt, "/tmp/valkey.sock");
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+### Executing commands
+
+The primary command interface is a `printf`-like function that takes a format string along with a variable number of arguments. This will construct a resp command and deliver it to the server.
+
+```c
+valkeyReply *reply = valkeyCommand(ctx, "INCRBY %s %d", "counter", 42);
+
+if (reply == NULL) {
+    fprintf(stderr, "Communication error: %s\n", c->err ? c->errstr : "Unknown error");
+} else if (reply->type == REDIS_REPLY_ERROR) {
```
bjosv, line 89:
```suggestion
} else if (reply->type == VALKEY_REPLY_ERROR) {
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+    fprintf(stderr, "Error response from server: %s\n", reply->str);
+} else if (reply->type != REDIS_REPLY_INTEGER) {
```
bjosv, line 91:
```suggestion
} else if (reply->type != VALKEY_REPLY_INTEGER) {
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+| VALKEY\_OPT\_NO\_PUSH\_AUTOFREE | Tells libvalkey to not install the default RESP3 PUSH handler (which just intercepts and frees the replies).  This is useful in situations where you want to process these messages in-band. |
+| VALKEY\_OPT\_NOAUTOFREEREPLIES | **ASYNC**: tells libvalkey not to automatically invoke `freeReplyObject` after executing the reply callback. |
+| VALKEY\_OPT\_NOAUTOFREE | **ASYNC**: Tells libvalkey not to automatically free the `valkeyAsyncContext` on connection/communication failure, but only if the user makes an explicit call to `valkeyAsyncDisconnect` or `valkeyAsyncFree` |
+
+### Executing commands
+
+The primary command interface is a `printf`-like function that takes a format string along with a variable number of arguments. This will construct a resp command and deliver it to the server.
```
bjosv, line 82:
```suggestion
The primary command interface is a `printf`-like function that takes a format string along with a variable number of arguments. This will construct a `RESP` command and deliver it to the server.
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+
+The `valkeyCommand` and `valkeyCommandArgv` functions return a `valkeyReply` on success and `NULL` in the event of a severe error (e.g. a communication failure with the server, out of memory condition, etc).
+
+If the reply is `NULL` you can inspect the nature of the error by querying `valkeyContext->err` for the error code and `valkeyContext->errstr` for a human readable error string.
+
+When a `valkeyReply` is returned, you should test the `valkeyReply->type` field to determine which kind of reply was received from the server. If for example there was an error in the command, this reply can be `VALKEY_REPLY_ERROR` and the specific error string will be in the `reply->str` member.
+
+### Reply types
+
+- `VALKEY_REPLY_ERROR` - An error reply. The error string is in `reply->str`
```
bjosv, line 127:
```suggestion
- `VALKEY_REPLY_ERROR` - An error reply. The error string is in `reply->str`.
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+- `VALKEY_REPLY_INTEGER` - An integer reply, which will be in `reply->integer`.
+- `VALKEY_REPLY_DOUBLE` - A double reply which will be in `reply->dval` as well as `reply->str`.
+- `VALKEY_REPLY_NIL` - a nil reply.
+- `VALKEY_REPLY_BOOL` - A boolean reply which will be in `reply->integer`.
+- `VALKEY_REPLY_BIGNUM` - As of yet unused, but the string would be in `reply->str`.
+- `VALKEY_REPLY_STRING` - A string reply which will be in `reply->str`.
+- `VALKEY_REPLY_VERB` - A verbatim string reply which will be in `reply->str` and who's type will be in `reply->vtype`.
+- `VALKEY_REPLY_ARRAY` - An array reply where each element is in `reply->element` with the number of elements in `reply->elements`.
+- `VALKEY_REPLY_MAP` - A map reply, which structurally looks just like `VALKEY_REPLY_ARRAY` only is meant to represent keys and values. As with an array reply you can access the elements with `reply->element` and `reply->elements`.
+- `VALKEY_REPLY_SET` - Another array-like reply representing a set (e.g. a reply from `SMEMBERS`). Access via `reply->elemeent` and `reply->elements`.
```
bjosv, line 138:
```suggestion
- `VALKEY_REPLY_SET` - Another array-like reply representing a set (e.g. a reply from `SMEMBERS`). Access via `reply->element` and `reply->elements`.
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+- `VALKEY_REPLY_ATTR` - An attribute reply. As of yet unused by valkey-server.
+- `VALKEY_REPLY_PUSH` - An out of band push reply. This is also array-like in nature.
+
+### Disconnecting / cleanup
+
+When libvalkey returns non-null `valkeyRepy` struts you are responsible for freeing them with `freeReplyObject`.  In order to disconnect and free the context simply call `valkeyFree`.
```
bjosv, line 144:
```suggestion
When libvalkey returns non-null `valkeyReply` structs you are responsible for freeing them with `freeReplyObject`.  In order to disconnect and free the context simply call `valkeyFree`.
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+// Error handling ...
+freeReplyObject(reply);
+
+// Disconnect and free context
+valkeyFree(ctx);
+```
+
+### Pipelining
+
+`valkeyCommand` and `valkeyCommandArgv` each make a round-trip to the server, by sending the command and then waiting for a reply`. Alternatively commands may be pipelined with the `valkeyAppendCommand` and `valkeyAppendCommandArgv` functions.
```
bjosv, line 157:
```suggestion
`valkeyCommand` and `valkeyCommandArgv` each make a round-trip to the server, by sending the command and then waiting for a reply. Alternatively commands may be pipelined with the `valkeyAppendCommand` and `valkeyAppendCommandArgv` functions.
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+- `VALKEY_ERR_IO` - A problem with the connection.
+- `VALKEY_ERR_EOF` - The server closed the connection.
+- `VALKEY_ERR_PROTOCOL` - There was an error parsing the reply.
+- `VALKEY_ERR_TIMEOUT` - A connect, read, or write timeout.
+- `VALKEY_ERR_OOM` - Out of memory.
+- `VALKEY_ERR_OTHER` - Some other error (check `c->errstr` for details).
+
+### Thread safety
+
+Libvalkey context structs are not thread safe. You should not attempt to share them between threads, unless you really know what you're doing.
```
bjosv, line 210:
```suggestion
Libvalkey context structs are **not** thread safe. You should not attempt to share them between threads, unless you really know what you're doing.
```
since there has been a lot of thread related issues  :)

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+
+### Reader configuration
+
+Libvalkey contexts have a few more mechanisms you can customize to your needs.
+
+#### Maximum input buffer size
+
+libvalkey uses a buffer to hold incoming bytes, which is typically restored to the configurable max buffer size (`16KB`) when it is empty. To avoid continually reallocating this buffer you can set the value higher, or to zero which means "no limit".
```
bjosv, line 218:
```suggestion
Libvalkey uses a buffer to hold incoming bytes, which is typically restored to the configurable max buffer size (`16KB`) when it is empty. To avoid continually reallocating this buffer you can set the value higher, or to zero which means "no limit".
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+```
+
+#### Allocator injection
+
+Internally libvalkey uses a layer of indirection from the standard allocation functions, by keeping a global structure with function pointers to the allocators we are going to use. By default they are just set to `malloc`, `calloc`, `realloc`, etc.
+
+These can be overridden like so
+
+```c
+valkeyAllocatorFuncs my_allocators = {
```
bjosv, line 270:
```suggestion
valkeyAllocFuncs my_allocators = {
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+
+```c
+valkeyResetAllocators();
+```
+
+**NOTE**: The `vk_calloc` function handles the case where `nmemb` * `size` would overflow a `size_t` and returns `NULL` in that case.
+
+## Asynchronous API
+
+Libvalkey also has an asynchronous API which supports a great many different event libraries. See the [examples](examples) directory for specific information about each individual event library.
```
bjosv, line 292:
```suggestion
Libvalkey also has an asynchronous API which supports a great many different event libraries. See the [examples](../examples) directory for specific information about each individual event library.
```

docs/standalone.md
```diff
@@ -0,0 +1,354 @@
...
+}
+
+int exec_some_commands(struct my_app_data *data) {
+    valkeyAsyncCommand(ac, my_incrby_callback, data, "INCRBY mykey %d", 42);
+    valkeyAsyncCommand(ac, my_get_callback, data, "GET %s", "mykey");
+}
+
+```
+
+See the [examples](../examples) directory for specific information on each supported event library.
```
bjosv, line 354: already mentioned above? but maybe is good to repeat

bjosv (commented):
Great work, looks nice. Just found some nits.
I guess I have to follow up with a cluster version, but now there is a structure that easily can be followed :)

michael-grunder:
Thanks, I think I got all the changes.

bjosv (approved):
LGTM!

PR #66, opened 2024-08-12
Attempt to fix timing test CI failure.

michael-grunder:
* Reduce sleep time from 3s to 1s. No need to sleep so long.
* Relax tolerances so we can be pretty sure the server has awakened when we make the next attempt.

### DIFF
--- a/tests/client_test.c
+++ b/tests/client_test.c
@@ -94,6 +94,7 @@ static void millisleep(int ms) {
     Sleep(ms);
 #else
     struct timespec ts = { ms / 1000, (ms % 1000) * 1000000 };
+
     nanosleep(&ts, NULL);
 #endif
 }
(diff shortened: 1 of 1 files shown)


tests/client_test.c
```diff
@@ -94,6 +94,7 @@ static void millisleep(int ms) {
     Sleep(ms);
 #else
     struct timespec ts = { ms / 1000, (ms % 1000) * 1000000 };
+
```
bjosv, line 97: Can be removed

tests/client_test.c
```diff
@@ -1359,7 +1356,7 @@ static void test_blocking_connection_timeouts(struct config config) {
         freeReplyObject(reply);
 
         // wait for the DEBUG SLEEP to complete so that the server is unblocked for the following tests
-        millisleep(3000);
+        millisleep(1100);
```
bjosv, line 1359:
```suggestion
        millisleep(1500);
```

bjosv (approved):
Great finding.
I ran the target separately and your change seems to fix it.
When running the tests 20 times in a row, within the same job, it failed 10 of 20 without the change.
With this fix it failed 1 of 20. This is good enough, but extending the time just a bit more to 1.5s would be ok for me. Then it was 0 of 20 fails when I ran a couple of times.

PR #67, opened 2024-08-13
Converge the warning-flags in Makefile and CMake builds    

bjosv:
Enable the same warnings in CMake as used in the Makefile (including `-Werror`).
Enable `-pedantic` in Makefile, as in CMake, to `issue all the warnings demanded by strict ISO C `.

Also fixes:
- Correcting -Wwrite-strings warnings in cluster tests (warns about the deprecated conversion from string constants to `char*`).
- Use the correct libevent2 type `evutil_socket_t` instead of `int` when implementing the libevent2 callback function in our tests. On Windows evutil_socket_t is an intptr_t (long long int) which gives a warning (and error with gcc-14).
    (Required by #62)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,8 +22,6 @@ jobs:
       - name: Build using cmake
         env:
           EXTRA_CMAKE_OPTS: -DENABLE_EXAMPLES:BOOL=ON -DENABLE_SSL:BOOL=ON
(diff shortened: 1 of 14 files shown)


bjosv:
Not sure if adding `-pedantic` in the Makefile gives any problems? like for maintainers of packages

michael-grunder (approved):
LGTM

PR #70, opened 2024-08-19
Add new APIs for variadic cluster commands

bjosv:
- valkeyClustervCommandToNode()
- valkeyClustervAppendCommandToNode()

From https://github.com/Nordix/hiredis-cluster/pull/231

### DIFF
--- a/include/valkey/valkeycluster.h
+++ b/include/valkey/valkeycluster.h
@@ -226,6 +226,9 @@ void *valkeyClusterCommandToNode(valkeyClusterContext *cc,
 /* Variadic using va_list */
 void *valkeyClustervCommand(valkeyClusterContext *cc, const char *format,
                             va_list ap);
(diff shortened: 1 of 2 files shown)


zuiderkwast (approved):
Yes, these are just missing variants.

The spellcheck error is not introduced here. Should be fixed in another PR. (Just add "valkey" to the wordlist?)

bjosv:
Yes, its seems to complain about following text in standalone.md.
`See [include/valkey.h](include/valkey.h) for more details`
Weird though..possible a simple fix, or alternatively we replace spellcheck with `typos` as in the valkey repo

bjosv:
Adding the word seem to be enough this time

zuiderkwast:
It may be the case that the spellchecker ignores text within backticks. I have a vague memory of this. If so, we could silence this kind of errors by placing backticks around the file name.

PR #73, opened 2024-08-19
Don't initiate new connections during a cluster client disconnect (async API)

bjosv:
- No slot map updates during a cluster client disconnect.
- Don't follow redirects when a cluster client disconnects.
- Don't accept new commands when a cluster client disconnects.

This is a corresponding change to https://github.com/Nordix/hiredis-cluster/pull/225 .

### DIFF
--- a/include/valkey/valkeycluster.h
+++ b/include/valkey/valkeycluster.h
@@ -58,6 +58,9 @@
 /* Flag to enable routing table updates using the command 'cluster slots'.
  * Default is the 'cluster nodes' command. */
 #define VALKEYCLUSTER_FLAG_ROUTE_USE_SLOTS 0x4000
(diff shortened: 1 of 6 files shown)


zuiderkwast (approved):
LGTM

Among the untouched lines, I noticed the below version definitions. We don't need those, do we? Valkey cluster and valkey is just one project with one version?

```C
#define LIBVALKEYCLUSTER_MAJOR 0
#define LIBVALKEYCLUSTER_MINOR 14
#define LIBVALKEYCLUSTER_PATCH 0
#define LIBVALKEYCLUSTER_SONAME 0.14
```

bjosv:
> Among the untouched lines, I noticed the below version definitions. We don't need those, do we? 

Ah, great finding. Will remove.

PR #75, opened 2024-08-22
Add file COPYING

bjosv:
Keep the information from `hiredis` and `hiredis-cluster` and added this projects license on top.
Used https://github.com/valkey-io/valkey/blob/unstable/COPYING as a template for this.


zuiderkwast (commented):
License 2 and license 3 are identical, right? We can merge the copyright names and call this license 2.

bjosv:
They are equal, only the copyright-lines differs, i.e added in hiredis-cluster

michael-grunder (approved):

PR #78, opened 2024-08-26
Small changes after formatting

mkmkme:
The first commit adds `.git-blame-ignore-revs` that lists the formatting commit so it would be omitted in `git blame`.

The second commit mentions `clang-format` version in README.


bjosv (approved):
LGTM. We could have a CONTRIBUTING.md file, but since we have some info in the readme already, lets start here.

zuiderkwast (approved):
LGTM

PR #79, opened 2024-08-27
Update version

mkmkme:
This PR contains two commits:
1. Make `LIBVALKEY_SONAME` a string to remove `clang-format off` block and to make it available from the code in case it's needed
2. Update version to 0.1.0 and make `LIBVALKEY_SONAME` of a format `${MAJOR}.${MINOR}`


bjosv (approved):
LGTM.
If we skip `-dev`, then I guess we don't need to handle `"`
But now it will work fine when/if we do add it..

PR #81, opened 2024-08-28
Enable release-drafter in CI

bjosv:
Enable the release drafter since it was used in hiredis,
and other clients under https://github.com/valkey-io seems keep using it.

### DIFF
--- a/.github/release-drafter-config.yml
+++ b/.github/release-drafter-config.yml
@@ -18,7 +18,6 @@ categories:
   - title: 'Breaking Changes'
     labels:
       - 'breakingchange'
-
   - title: '🧪 Experimental Features'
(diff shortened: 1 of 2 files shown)


bjosv:
I have no experience in using the release-drafter, so I don't really know if its helpful or if we should write the changelog/release notes ourselves..

michael-grunder (approved):

michael-grunder:
I like it personally. My biggest issue was forgetting to apply appropriate labels before merging but that's user error 😅

bjosv:
Ah, then we use it. I'll just have too look into how its works. Hopefully its possible to apply the labels retroactively before we tag and release.

PR #84, opened 2024-09-02
Remove unused parsing of replicas/slaves in cluster API

bjosv:
The option to parse replicas are disabled by default and when enabled its not used for anything.


bjosv:
Will be refactored in other PR.

PR #85, opened 2024-09-03
Add initial migration guide

bjosv:
https://github.com/bjosv/libvalkey/blob/add-migration-guide/docs/migration-guide.md

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -11,7 +11,12 @@ The general actions needed are:
 
 ## Migrating from `hiredis` v1.2.0
 
-No version specific action needed.
+The type `sds` is removed from the public API.
+
+### Removed API functions
+
+* `valkeyFormatSdsCommandArgv` removed from API. Can be replaced with `valkeyFormatCommandArgv`.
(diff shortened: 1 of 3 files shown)


michael-grunder (approved):

docs/migration-guide.md
```diff
@@ -0,0 +1,46 @@
...
+  All `libvalkey` headers are now found under `include/valkey/`.
+
+## Migrating from `hiredis` v1.2.0
+
+The type `sds` is removed from the public API.
+
+### Removed API functions
+
+* `valkeyFormatSdsCommandArgv` removed from API. Can be replaced with `valkeyFormatCommandArgv`.
+* `valkeyFreeSdsCommand` removed since the `sds` type is for internal use only.
```
zuiderkwast, line 19: Hiredis 1.2.0 doesn't have a function called `valkeyFormatSdsCommandArgv`.

```suggestion
* `redisFormatSdsCommandArgv` removed from API. Can be replaced with `valkeyFormatCommandArgv`.
* `redisFreeSdsCommand` removed since the `sds` type is for internal use only.
```

bjosv: of course! Good finding

PR #87, opened 2024-09-04
Remove `dict` from user-facing API

### LINKED ISSUE #44: Move dict.h from include/ to src/
We should move files like dict.h from `include/` to `src/` to avoid exposing internal functions in the public API.

bjosv:
This PR moves `dict.h` from `include/valkey/` to `src/` which mainly affects `valkeyClusterNodeIterator`.

By making the type `valkeyClusterNodeIterator` opaque we can hide its content and remove the exposure of the `dictIterator`.

Fixes #44


zuiderkwast:
If we want a stack allocated opaque struct, we need to expose the size but we can define it like

```C
struct valkeyClusterNodeIterator {
    char opaque_data[NODE_ITERATOR_SIZE];
};
```

With an appropriate comment about not relying on the size...

src/valkeycluster.c
```diff
@@ -3687,6 +3687,20 @@ void valkeyClusterAsyncFree(valkeyClusterAsyncContext *acc) {
...
 
+struct valkeyClusterNodeIterator {
+    valkeyClusterContext *cc;
+    uint64_t route_version;
+    int retries_left;
+    dictIterator di;
+};
+
+/* Get an initated iterator for iterating over current cluster nodes */
+valkeyClusterNodeIterator *valkeyClusterGetNodeIterator(valkeyClusterContext *cc) {
```
zuiderkwast, line 3698: Shall we call this create instead of get? A getter doesn't normally allocate anything.

```suggestion
valkeyClusterNodeIterator *valkeyClusterCreateNodeIterator(valkeyClusterContext *cc) {
```

bjosv:
Not sure if this would be fragile when it comes to using it on new/other platforms with different packing of structs?
WDYT?

src/vkutil.h
```diff
@@ -36,6 +36,9 @@
 #include <stdint.h>
 #include <sys/types.h>
 
+/* Static assert macro for C99. */
+#define vk_static_assert(cond) extern char vk_static_assert[sizeof(char[1 - 2 * !(cond)])]
```
michael-grunder, line 40:
```suggestion
#define vk_static_assert(cond) extern char vk_static_assert[sizeof(char[(cond) ? 1 : -1])]
```

michael-grunder (approved):

src/vkutil.h
michael-grunder: Just because I'm slow and the `1 - 2 * !(cond)` took me too long to parse :smile:

zuiderkwast (approved):

PR #88, opened 2024-09-05
Remove `valkey` from filenames and update header guards

bjosv:
Since we have change the file structure for public headers we can remove `valkey` from their filenames.
As example, we will have `#include <valkey/ssl.h>` instead of `#include <valkey/valkey_ssl.h>`.
- `valkeycluster.h` -> `cluster.h`
- `valkeycluster_ssl.h` -> `cluster_ssl.h`
- `valkey_rdma.h` -> `rdma.h`
- `valkey_ssl.h` -> `ssl.h`

This PR also includes:
- The header guards are updated to follow the common rule: `<PROJ>_<PATH>_<FILENAME>_H`
- Renaming of 2 cluster source files, removing `valkey` from their names.
- Make sure `ssl.h` can be included before `valkey.h` (required by clang-format after the rename).


tests/client_test.c
```diff
@@ -26,10 +26,10 @@
 #include <math.h>
 #include <signal.h>
 #ifdef VALKEY_TEST_SSL
-#include "valkey_ssl.h"
+#include "ssl.h"
```
zuiderkwast, line 29: Minor:

In the examples we include `"valkey/ssl.h"` but in tests we write just `"ssl.h"`, because we use different compiler flags?

If we want the tests to be useful as examples, then maybe we should change them to `"valkey/ssl.h"` style too?

zuiderkwast (approved):

tests/client_test.c
bjosv: The examples are only [built in CI](https://github.com/valkey-io/libvalkey/blob/main/.github/workflows/ci.yml#L102) by first installing libvalkey on the runner, and then build the examples using the installed headers and libs. So, here we use `<valkey/ssl.h>`. This is a verification job for `make install`.

But its a good idea, I'll take a stab at that in a separate PR.
The MakeList/Makefile needs a slight change and there will be a bunch of include updates.

michael-grunder (approved):

tests/client_test.c
zuiderkwast: OK, or just ignore this if it's non-trivial.

PR #91, opened 2024-09-10
Cleanup of cluster free functions

bjosv:
Remove special handling when freeing the nodes dict.
The problem is now prevented by the newly added flag `VALKEYCLUSTER_FLAG_DISCONNECTING` (#73).

Scrub the `valkeyClusterContext` with `0xff`'s before freeing to help us detect use-after-free.
This is similar to how `valkeyContext` is [freed](https://github.com/valkey-io/libvalkey/blob/f888a144b7d5863c1efac033f912a7ca2963b707/src/valkey.c#L758) and added for consistent behavior.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1364,7 +1364,6 @@ valkeyClusterContext *valkeyClusterContextInit(void) {
 }
 
 void valkeyClusterFree(valkeyClusterContext *cc) {
-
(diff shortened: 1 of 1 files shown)


bjosv:
valkeyClusterFree can be called by the user too, like valkeyFree can.

PR #92, opened 2024-09-10
Add sanitizers to CI and fix a found test issue

bjosv:
Update the CI workflow `ubuntu`:
- Build using the now available GCC 14.
- Build for Release only to use less CPU time.
- Enable sanitizers.

The PR includes:
- Fix `global-buffer-overflow` in client_test
  ..where the string length of "*1\r\n,3.14159265358979323846\r\n" is 29.


michael-grunder (approved):
Good catch on the 31 -> 29 multibulk

PR #93, opened 2024-09-11
Cleanup of internal cluster error functions

bjosv:
Add internal convenience functions to clear errors, and cleanup the internal `SetError` functions.
The error string for `VALKEY_ERR_IO` is set by the standalone API, so removing the handling for it in the cluster part.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -200,23 +200,20 @@ static unsigned int keyHashSlot(char *key, int keylen) {
 
 static void valkeyClusterSetError(valkeyClusterContext *cc, int type,
                                   const char *str) {
-    size_t len;
-
-    if (cc == NULL) {
-        return;
-    }
-
     cc->err = type;
-    if (str != NULL) {
-        len = strlen(str);
+
+    assert(str != NULL);
+    if (str != NULL && str != cc->errstr) {
+        size_t len = strlen(str);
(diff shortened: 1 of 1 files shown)


src/cluster.c
```diff
@@ -200,25 +200,24 @@ static unsigned int keyHashSlot(char *key, int keylen) {
...
+        size_t len = strlen(str);
         len = len < (sizeof(cc->errstr) - 1) ? len : (sizeof(cc->errstr) - 1);
         memcpy(cc->errstr, str, len);
         cc->errstr[len] = '\0';
-    } else {
-        /* Only VALKEY_ERR_IO may lack a description! */
-        assert(type == VALKEY_ERR_IO);
-        strerror_r(errno, cc->errstr, sizeof(cc->errstr));
+    } else if (str == NULL) { /* No known usecase. Clear previous errstr. */
+        cc->errstr[0] = '\0';
```
zuiderkwast, line 212: If this is never called, we should probably do an assert instead, right?

    assert(str != NULL);

bjosv: I have gone through a bit more, and it should be safe to assert.
Keeping the if-check part in case of NDEBUG, which is set by CMake in release builds.

zuiderkwast (approved):

PR #95, opened 2024-09-13
CI updates

bjosv:
- Refactor and move the CI step to build using make.
 Run the tests under valgrind on Ubuntu only, which should be enough.

- Add a CMake build with type=Debug to make sure building without `-DNDEBUG` is fine.


michael-grunder (approved):
LGTM

PR #103, opened 2024-09-25
Replace the term SSL with TLS

bjosv:
This PR replaces the use of the term `SSL` with `TLS`, as done in `valkey`, and is mainly a search and replace but grouped in related chunks. The change can be reviewed in separate steps, each step is an own commit with a green CI.
But still, sorry about the size..

The commits are:
1. `Remove forward declaration of non-existing struct valkeySsl`
  This is a on-line cleanup.
2. `Rename files to use 'tls' instead of 'ssl'` in own commit.
3. `Updates needed after file renaming`
   Making the build pass after renaming files.
4. `Rename ssl library and examples`
   Use the name `libvalkey_tls` instead of `libavalkey_ssl`, update CMakeLists/Makefile and rename example binaries.
5. `Update API and functions to use tls instead of ssl`
  Rename public API, and internal  types and code.
7. `Update build and test flags`
    CMake builds now accepts `ENABLE_TLS` instead of `ENABLE_SSL`.
    Make builds now accepts `USE_TLS` instead of `USE_SSL`.
    Standalone test.sh now accepts `TEST_TLS` instead of `TEST_SSL`.
8. `Update docs and migration guide`


zuiderkwast (approved):
LGTM.

Since we're rename everything anyway, now is a good time to rename this as well.

michael-grunder (approved):
Only downside is people having to build slightly differently when migrating but that's very minor.

bjosv:
> Only downside is people having to build slightly differently when migrating but that's very minor.

I guess so, this is an additional change to their builds besides fiddling with new header paths and library names.

PR #108, opened 2024-10-06
Use `-std=c99` when building with CMake

bjosv:
Disable the GNU extensions in CMake, which is ON by default for some reason.
This enables `-std=c99` instead of `-std=gnu99`, same as in the Makefile.

Enable GNU extensions where required in examples and some tests.


michael-grunder (approved):
LGTM.

PR #112, opened 2024-10-09
Add unit tests for `CLUSTER NODES` parsing

### LINKED ISSUE #33: Improve tests of parsing `CLUSTER NODES` response
Add tests to cover slot map updates using `CLUSTER NODES`, but with [special slot entries](https://valkey.io/commands/cluster-nodes/#special-slot-entries).

Some existing tests are:
https://github.com/valkey-io/libvalkey/blob/main/tests/scripts/moved-redirect-using-cluster-nodes-test.sh
https://github.com/valkey-io/libvalkey/blob/main/tests/scripts/connect-error-using-cluster-nodes-test.sh
https://github.com/valkey-io/libvalkey/blob/main/tests/scripts/moved-redirect-using-cluster-nodes-test.sh

bjosv:
Includes OOM testing of replica parsing and fixes for memory issues when parsing fails.
    

Fixes #33


src/cluster.c
```diff
@@ -1042,8 +1041,19 @@ static dict *parse_cluster_nodes(valkeyClusterContext *cc, valkeyReply *reply) {
 error:
     sdsfreesplitres(part, count_part);
     sdsfreesplitres(slot_start_end, count_slot_start_end);
+    if (nodes_name != NULL) {
+        /* Only free parsed replicas since the `nodes` dict owns primary nodes. */
```
zuiderkwast, line 1045: This is a little odd and maybe fragile.

What about the corner case primaries without any slots? Does the `nodes` dict own them too?

tests/ct_out_of_memory_handling.c
```diff
@@ -170,14 +171,14 @@ void test_alloc_failure_handling(void) {
 
     // Connect
     {
-        for (int i = 0; i < 128; ++i) {
+        for (int i = 0; i < 148; ++i) {
```
zuiderkwast, line 174: Because the stack-allocated iterator uses fewer allocations?

zuiderkwast (approved):
LGTM

src/cluster.c
bjosv: It is a bit odd.
The `nodes` dict owns all `valkeyClusterNodes` after the parsing is complete, these are nodes with-or-without slots. This dict holds key=address, value=primary node, and each primary owns a list of replicas.
Deleting the `nodes` dict will delete all its objects, primaries and their known replicas, when there is an error while parsing.

But then there is the temporary `nodes_name` dict which holds references to all currently parsed cluster nodes.
The found issue is that this dict is the actual owner of an replica node, that has a primary, which is not yet parsed.
So, these are the replica nodes that needs to be deleted while the parsing fails.

I have a coming PR where the cluster nodes parsing is refactored.

tests/ct_out_of_memory_handling.c
bjosv: I added the config to also parse replicas in this testcase, This required a couple of more allocations to succeed.

PR #119, opened 2024-10-14
 Refactor the internal function `parse_cluster_nodes`

bjosv:
This is a 2 step/2 commit refactor PR.

* Refactor the internal function `parse_cluster_nodes`.
Parse the `CLUSTER NODES` reply by a simpler string search using `strchr` and string modification instead of using `sdssplitlen`, which allocates memory for each element.
Parse each line once using a new function `parse_cluster_nodes_line`, which returns a `valkeyClusterNode`.
Replicas are only parsed when configured, like legacy.

* Refactor the replica handling in CLUSTER NODES parsing.
Store parsed replicas in a separate dict during parsing, and move them to their primary when all lines are parsed.
This dict owns the memory for replicas until moved to the related primary node.
Previously the dict contained references to both replicas and primaries and a bit harder to handle.
The function `cluster_master_slave_mapping_with_name` is removed and its dual purpose handling is replaced by functions `store_replica_node` and `move_replica_nodes`.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -720,7 +720,8 @@ static int store_replica_node(dict *replicas, char *primary_id, valkeyClusterNod
     return VALKEY_OK;
 }
 
-/* Move parsed replica nodes from the collection to related primary. */
+/* Move parsed replica nodes to its primary node, which holds a list of replica
+ * nodes. The `replicas` dict shall contain nodes with primary_id as key. */
 static int move_replica_nodes(dict *replicas, dict *nodes) {
     if (replicas == NULL)
         return VALKEY_OK;
@@ -770,15 +771,19 @@ static int move_replica_nodes(dict *replicas, dict *nodes) {
     return VALKEY_OK;
 }
 
-/* Parse a node from a single CLUSTER NODES line. Only parse primary nodes if
- * the 'replica_master_id' argument is NULL, otherwise replicas are parsed and
- * its master_id is given via 'replica_master_id'. */
+/* Parse a node from a single CLUSTER NODES line. Returns an allocated
+ * valkeyClusterNode as a pointer in `parsed_node`.
+ * Only parse primary nodes if the `parsed_primary_id` argument is NULL,
+ * otherwise replicas are also parsed and its primary_id is returned by pointer
+ * via 'parsed_primary_id'. */
 static int parse_cluster_nodes_line(valkeyClusterContext *cc, char *line,
                                     valkeyClusterNode **parsed_node, char **parsed_primary_id) {
     char *p, *id = NULL, *addr = NULL, *flags = NULL, *primary_id = NULL,
              *link_state = NULL, *slots = NULL;
+    /* Find required fields and keep a pointer to each field:
+     * <id> <addr> <flags> <primary_id> <ping-sent> <pong-recv> <config-epoch> <link-state> [<slot> ...]
+     */
     // clang-format off
-    /* Find required fields. */
     int i = 0;
     while ((p = strchr(line, ' ')) != NULL) {
         *p = '\0';
@@ -800,7 +805,8 @@ static int parse_cluster_nodes_line(valkeyClusterContext *cc, char *line,
         return VALKEY_ERR;
     }
 
-    /* Parse flags. */
+    /* Parse flags, a comma separated list of following flags:
+     * myself, master, slave, fail?, fail, handshake, noaddr, nofailover, noflags. */
     uint8_t role = VALKEY_ROLE_NULL;
     while (*flags != '\0') {
         if ((p = strchr(flags, ',')) != NULL)
@@ -837,8 +843,8 @@ static int parse_cluster_nodes_line(valkeyClusterContext *cc, char *line,
     if (node->name == NULL)
         goto oom;
 
-    /* Handle address field <ip:port@cport...>
-     * Remove @cport... since addr is used as a dict key which should be <ip>:<port> */
+    /* Parse the address field: <ip:port@cport[,hostname]>
+     * Remove @cport.. to get <ip>:<port> which is our dict key. */
     if ((p = strchr(addr, PORT_CPORT_SEPARATOR)) != NULL) {
         *p = '\0';
     }
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -780,8 +780,8 @@ static int move_replica_nodes(dict *replicas, dict *nodes) {
  * the 'replica_master_id' argument is NULL, otherwise replicas are parsed and
  * its master_id is given via 'replica_master_id'. */
 static int parse_cluster_nodes_line(valkeyClusterContext *cc, char *line,
-                                    valkeyClusterNode **parsed_node, char **replica_master_id) {
-    char *p, *id = NULL, *addr = NULL, *flags = NULL, *master_id = NULL,
+                                    valkeyClusterNode **parsed_node, char **parsed_primary_id) {
+    char *p, *id = NULL, *addr = NULL, *flags = NULL, *primary_id = NULL,
              *link_state = NULL, *slots = NULL;
     // clang-format off
     /* Find required fields. */
@@ -792,7 +792,7 @@ static int parse_cluster_nodes_line(valkeyClusterContext *cc, char *line,
             case 0: id = line; break;
             case 1: addr = line; break;
             case 2: flags = line; break;
-            case 3: master_id = line; break;
+            case 3: primary_id = line; break;
             case 7: link_state = line; break;
         }
         line = p + 1; /* Start of next field. */
@@ -829,7 +829,7 @@ static int parse_cluster_nodes_line(valkeyClusterContext *cc, char *line,
     }
 
     /* Only parse replicas when requested. */
-    if (role == VALKEY_ROLE_SLAVE && replica_master_id == NULL) {
+    if (role == VALKEY_ROLE_SLAVE && parsed_primary_id == NULL) {
         *parsed_node = NULL;
         return VALKEY_OK;
     }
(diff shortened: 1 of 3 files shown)


src/cluster.c
```diff
@@ -852,176 +689,306 @@ static dict *parse_cluster_slots(valkeyClusterContext *cc, valkeyReply *reply) {
...
+        valkeyClusterSetError(cc, VALKEY_ERR_OTHER, "Mandatory fields missing");
+        return VALKEY_ERR;
+    }
+
+    /* Parse flags. */
+    uint8_t role = VALKEY_ROLE_NULL;
+    while (*flags != '\0') {
+        if ((p = strchr(flags, ',')) != NULL)
+            *p = '\0';
+        if (memcmp(flags, "master", 6) == 0) {
```
zuiderkwast, line 814: Can you add an example line in a comment, maybe with the names of the different parts underneath? (flags, role, etc.) It would make it easier to read code like this.

```
a8b79a8b798ab master 192.0.0.2:123@456,example.com,foo=bar ...
node-id       role   ip        port cport hostname flags   ...
```

src/cluster.c
```diff
@@ -852,176 +689,306 @@ static dict *parse_cluster_slots(valkeyClusterContext *cc, valkeyReply *reply) {
...
-    int len;
-
-    if (reply->type != VALKEY_REPLY_STRING) {
-        valkeyClusterSetError(cc, VALKEY_ERR_OTHER, "Unexpected reply type");
-        goto error;
+/* Store a parsed replica node in given dict using the primary_id as key.
+ * Additional replicas for a primary are stored within the first replica. */
+static int store_replica_node(dict **replicas, char *primary_id, valkeyClusterNode *node) {
+    if (*replicas == NULL) {
+        *replicas = dictCreate(&clusterNodesDictType, NULL);
```
zuiderkwast, line 696: This is a slightly odd API, but I guess there's some point in avoiding creating the dict if it isn't needed.... (Create it before we call this function doesn't make sense?)

Can we call it `replicas_ref` or something that indicates it's not the replicas dict itself but a reference or pointer to it?

src/cluster.c
```diff
@@ -852,176 +689,306 @@ static dict *parse_cluster_slots(valkeyClusterContext *cc, valkeyReply *reply) {
...
-    line_start = start;
+    return VALKEY_OK;
+}
 
-    for (pos = start; pos < end; pos++) {
-        if (*pos == '\n') {
-            line_end = pos - 1;
-            len = line_end - line_start;
+/* Move parsed replica nodes from the collection to related primary. */
+static int move_replica_nodes(dict *replicas, dict *nodes) {
```
zuiderkwast, line 730: From the comment and the names here, it's not entirely clear to me what this does.

"The related primaries" = `nodes`?

Why are they dicts? In the context where they're used, it probably makes sense, but here I'm wondering if the node is the key or the value in each dict...

src/cluster.c
```diff
@@ -852,176 +689,306 @@ static dict *parse_cluster_slots(valkeyClusterContext *cc, valkeyReply *reply) {
...
-            // if the address string starts with ":0", skip this node.
-            if (sdslen(part[1]) >= 2 && memcmp(part[1], ":0", 2) == 0) {
-                sdsfreesplitres(part, count_part);
-                count_part = 0;
-                part = NULL;
+/* Parse a node from a single CLUSTER NODES line. Only parse primary nodes if
+ * the 'replica_master_id' argument is NULL, otherwise replicas are parsed and
+ * its master_id is given via 'replica_master_id'. */
+static int parse_cluster_nodes_line(valkeyClusterContext *cc, char *line,
+                                    valkeyClusterNode **parsed_node, char **replica_master_id) {
```
zuiderkwast, line 783: "Given" usually means input parameter, as in given by the caller, but this is returned by pointer, right? Or maybe I don't really understand what you mean here.

If it is return-by-pointer, I usually document is as something like "replica_master_id is pointed to the master id", or just "returned by pointer".

Please document `parsed_node` too. (It is pointed to the parsed node, right?)

zuiderkwast (commented):
I didn't review everything very carefully, but I believe you just refactored existing logic and we have tests covering it, so it's fine by me.

We use the words master, slave, primary, replica a little bit mixed everywhere, in macros, functions, struct fieds, variables and comments. I guess we should strive towards the new terminology, but not change all in the same PR?

src/cluster.c
bjosv: Added comment about which fields are parsed/expected

src/cluster.c
bjosv: Changed it. Now creates the dict before calling this function.

src/cluster.c
bjosv: Added comments with some more details.

bjosv:
> I didn't review everything very carefully, but I believe you just refactored existing logic and we have tests covering it, so it's fine by me.
>
Yes, the newly added testcases in ut_slotmap_update.c should cover this change.

> We use the words master, slave, primary, replica a little bit mixed everywhere, in macros, functions, struct fieds, variables and comments. I guess we should strive towards the new terminology, but not change all in the same PR?

The plan was to change the define and struct in a separate PR since it also affects CLUSTER SLOTS parsing, but there was some new code in this PR that was a bit inconsistent, this is now changed.

src/cluster.c
```diff
@@ -852,176 +689,312 @@ static dict *parse_cluster_slots(valkeyClusterContext *cc, valkeyReply *reply) {
...
+    sdsfree(key);
+    valkeyClusterNode *n = dictGetEntryVal(de);
+    if (n->slaves == NULL) {
+        n->slaves = listCreate();
+        if (n->slaves == NULL)
+            return VALKEY_ERR;
+        n->slaves->free = listClusterNodeDestructor;
     }
+    if (listAddNodeTail(n->slaves, node) == NULL)
+        return VALKEY_ERR;
```
zuiderkwast, line 718: This seems very weird.

Should the replicas be added under `n->slaves` on the primary?

From what I can read here, `n` is a replica and the other replicas are stored under `n->slaves`.

src/cluster.c
```diff
@@ -852,176 +689,312 @@ static dict *parse_cluster_slots(valkeyClusterContext *cc, valkeyReply *reply) {
...
-    char *role;
-    int role_len;
-    int slot_start, slot_end, slot_ranges_found = 0;
-    sds *part = NULL, *slot_start_end = NULL;
-    int count_part = 0, count_slot_start_end = 0;
-    int k;
-    int len;
+/* Store a parsed replica node in given dict using the primary_id as key.
+ * Additional replicas for a primary are stored within the first replica. */
+static int store_replica_node(dict *replicas, char *primary_id, valkeyClusterNode *node) {
```
zuiderkwast, line 694: Is it written anywhere that the `replicas` dict maps primary-id to a replica node, where the other replicas are added as slaves to it..? (which is a very weird way to store it; see other comment)

bjosv: I first kept the logic how to hold replica nodes while parsing, but it was weird, I have now changed it.
I have now added a new dictType to hold replica nodes while parsing.

The dict now uses primary_id as key, which points to a list of replica nodes.
Since the order in `cluster nodes` is unknown; when all nodes have been parsed the list of replicas for a given primary is moved to `primary->slaves`. (to be renamed)

src/cluster.c
```diff
@@ -852,23 +692,231 @@ static dict *parse_cluster_slots(valkeyClusterContext *cc, valkeyReply *reply) {
...
+    char *p, *id = NULL, *addr = NULL, *flags = NULL, *primary_id = NULL,
+             *link_state = NULL, *slots = NULL;
+    /* Find required fields and keep a pointer to each field:
+     * <id> <addr> <flags> <primary_id> <ping-sent> <pong-recv> <config-epoch> <link-state> [<slot> ...]
+     */
+    // clang-format off
+    int i = 0;
+    while ((p = strchr(line, ' ')) != NULL) {
+        *p = '\0';
+        switch(i++){
```
zuiderkwast, line 768: Clang-format didn't mention this?

```suggestion
        switch (i++){
```

src/cluster.c
```diff
@@ -852,23 +692,231 @@ static dict *parse_cluster_slots(valkeyClusterContext *cc, valkeyReply *reply) {
...
+    while ((de = dictNext(&di))) {
+        valkeyClusterNode *primary = dictGetEntryVal(de);
+
+        /* Move replica nodes related to this primary. */
+        dictEntry *der = dictFind(replicas, primary->name);
+        if (der != NULL) {
+            assert(primary->slaves == NULL);
+            /* Move replica list from replicas dict to nodes dict. */
+            primary->slaves = dictGetEntryVal(der);
+            dictSetHashVal(replicas, der, NULL);
```
zuiderkwast, line 746: I don't recognize the function name `dictSetHashVal`. It must be copied from a very old Redis version. :smile:

zuiderkwast (approved):
Great, this is much less weird.

src/cluster.c
```diff
@@ -852,23 +692,231 @@ static dict *parse_cluster_slots(valkeyClusterContext *cc, valkeyReply *reply) {
...
+    char *p, *id = NULL, *addr = NULL, *flags = NULL, *primary_id = NULL,
+             *link_state = NULL, *slots = NULL;
+    /* Find required fields and keep a pointer to each field:
+     * <id> <addr> <flags> <primary_id> <ping-sent> <pong-recv> <config-epoch> <link-state> [<slot> ...]
+     */
+    // clang-format off
+    int i = 0;
+    while ((p = strchr(line, ' ')) != NULL) {
+        *p = '\0';
+        switch (i++){
```
bjosv, line 768:
```suggestion
        switch (i++) {
```

src/cluster.c
bjosv: It's disabled here to get a compact switch, but then this was missed.. :smile:
I now also see that there should be an additional space before `{`..

PR #123, opened 2024-10-22
Create cluster nodes dict and request list when initiating valkeyClusterContext

bjosv:
It simplifies the code a bit when these are allocated at context initiation.


zuiderkwast (approved):
Looks reasonable. We never free it except in the release function, so there is no chance it can ever be NULL now? I didn't check that OOM test suite with the hard-coded numbers.

bjosv:
> We never free it except in the release function, so there is no chance it can ever be NULL now?

Exactly, we only free in the release function.

PR #137, opened 2024-12-04
Introduce configuration options in the cluster API

bjosv:
The **standalone** API uses the options struct `valkeyOptions` to allow users to configure how `libvalkey` set up the context.
This struct is used when calling `valkeyConnectWithOptions(valkeyOptions *opt)`, which returns a connected context with prepared command timeout values and so on.

The current **cluster** API, which in inherited from the project `hiredis-vip` and `hiredis-cluster`, has some similarities to the standalone API, but it does not provide a options struct and does not have the same "look and feel" as the standalone API.

This PR introduces a [`valkeyClusterOption`](https://github.com/bjosv/libvalkey/blob/refs/heads/clusterconfig/include/valkey/cluster.h#L186) that allows a user to set all options before connecting to the cluster using:
`valkeyClusterContext *cc = valkeyClusterConnectWithOptions(&options)`
or 
`valkeyClusterAsyncContext *acc = valkeyClusterAsyncConnectWithOptions(&options)` when using the async API.

This options struct enables users to set the initial nodes, callbacks, timeouts, options, TLS- and event-adapters,
instead of using a wide range of `valkeyClusterSetOptionXxx` functions on a `valkeyClusterContext` (even when they use async contexts `valkeyClusterAsyncContext`...).
There are still options that are allowed to be configured during runtime like`valkeyClusterSetOptionTimeout`, but most other options requires a reconnect.

The async cluster API will no longer require that users calls the blocking cluster API using `valkeyClusterConnect2(acc->cc)`.
We now provide a new option `VALKEY_OPT_BLOCKING_INITIAL_UPDATE` to enable a blocking slotmap update after an initial connect using the `valkeyClusterAsyncConnectWithOptions` API. This will allow us to avoid `acc->cc` usages and we should be able to hide the `cc` in the `acc` struct.

All testcases are updated to use the new API and this should give a feeling how it works,
alternatively look at changes in `include/valkey/cluster.h`.


include/valkey/adapters/libevent.h
```diff
@@ -192,4 +192,12 @@ static int valkeyClusterLibeventAttach(valkeyClusterAsyncContext *acc,
...
     return VALKEY_OK;
 }
+
+/* Helper macro to initialize options. */
+#define VALKEY_CLUSTER_OPTIONS_SET_ADAPTER_LIBEVENT(opts, event_base) \
+    do {                                                              \
+        (opts)->attach_fn = valkeyLibeventAttachAdapter;              \
+        (opts)->attach_data = event_base;                             \
+    } while (0)
+
```
zuiderkwast, line 202: Why are these not static functions, like valkeyClusterLibeventAttach and others?

include/valkey/cluster_tls.h
```diff
@@ -43,6 +43,13 @@ extern "C" {
 int valkeyClusterSetOptionEnableTLS(valkeyClusterContext *cc,
                                     valkeyTLSContext *tls);
 
+/* Helper macro to initialize options. */
+#define VALKEY_CLUSTER_OPTIONS_SET_SSL(opts, tls_)           \
+    do {                                                     \
+        (opts)->tls = tls_;                                  \
+        (opts)->tls_init_fn = &valkeyInitiateTLSWithContext; \
+    } while (0)
+
```
zuiderkwast, line 52: Why not a function?

zuiderkwast (commented):
The API looks good in general!

include/valkey/cluster.h
```diff
@@ -168,21 +168,23 @@ typedef struct {
...
     int (*attach_fn)(valkeyAsyncContext *ac, void *attach_data);
     void *attach_data;
 
     /* Async API callbacks. */
-    valkeyConnectCallback *onConnect;
-    valkeyConnectCallbackNC *onConnectNC;
-    valkeyDisconnectCallback *onDisconnect;
+    valkeyConnectCallback *async_connect_cb;
+    valkeyConnectCallbackNC *async_connect_nc_cb; /* non-const callback */
+    valkeyDisconnectCallback *async_disconnect_cb;
```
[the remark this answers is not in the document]
bjosv: Good idea. Any thoughts or objections @michael-grunder ?

The const was first removed in hiredis:
```
    Drop `const` on redisAsyncContext in redisConnectCallback
    Since the callback is now re-entrant, it can call apis such as redisAsyncDisconnect()
```
but this change was reverted and the NC variant was added to the API instead:
```
    Maintain backward compatibiliy withour onConnect callback.
    Unfortunately, this is a breaking change we can't make until hiredis v2.0.0.
   
    This commit creates a separate callback member and corresponding
    function that allows us to use the new functionality, while maintaining
    our existing API for legacy code.
```

include/valkey/cluster_tls.h
bjosv: Changed to a function

include/valkey/adapters/libevent.h
bjosv: Changed to a function

docs/cluster.md
```diff
@@ -27,37 +29,57 @@ It is not intended as a complete reference. For that it's always best to refer t
...
-
-To be able to set options before any connection attempt is performed a `valkeyClusterContext` can be created using `valkeyClusterContextInit`.
-The context is where the known cluster state and the options are kept, which can be configured using functions like
-`valkeyClusterSetOptionAddNodes` to add one or many cluster node addresses,
-or `valkeyClusterSetOptionUsername` together with `valkeyClusterSetOptionPassword` to configure authentication and so on.
-See [`include/valkey/cluster.h`](../include/valkey/cluster.h) for more details.
+### Connection options
 
-The function `valkeyClusterConnect2` is provided to connect using the prepared `valkeyClusterContext`.
+There are a variety of options you can specify when connecting to a cluster, which are delivered via the `valkeyClusterOptions` helper struct.
```
zuiderkwast, line 56: Delivered? :thinking: specified maybe? or provided, configured, ...

Helper struct? Just struct.

docs/cluster.md
```diff
@@ -27,37 +29,57 @@ It is not intended as a complete reference. For that it's always best to refer t
...
+This includes information how to connect to the cluster and defining optional callbacks as well as other flags.
```
zuiderkwast, line 57: Missing word "about".

"callbacks and other flags" sounds like callbacks are flags. But it's not only flags. It's password and other things. Let's just say "and other options"?

```suggestion
This includes information about how to connect to the cluster and defining optional callbacks and other options.
```

docs/cluster.md
```diff
@@ -27,37 +29,57 @@ It is not intended as a complete reference. For that it's always best to refer t
...
-if (status == VALKEY_ERR) {
-    printf("Error: %s\n", cc->errstr);
-    // handle error
+valkeyClusterOptions opt = {0};
+opt.initial_nodes = "127.0.0.1:6379,127.0.0.1:6380"; // Addresses to initially connect to.
+opt.options = VALKEY_OPT_USE_CLUSTER_SLOTS;          // See available flags below.
+opt.password = "password"                            // Authentication; libvalkey sends the `AUTH` command.
+
+// There are helper functions to set some options.
+valkeyClusterOptionsEnableTLS(&opt, tlsCtx);         // Use TLS and a prepared `valkeyTLSContext` when connecting.
```
zuiderkwast, line 67: Why do we need this function to set TLS options? Can't you just set them on the opts object?

```c
opts.tls = tlsCtx;
```

docs/cluster.md
```diff
@@ -27,37 +29,57 @@ It is not intended as a complete reference. For that it's always best to refer t
...
+
+valkeyClusterContext *cc = valkeyClusterConnectWithOptions(&opt);
+if (cc == NULL || cc->err) {
+    fprintf(stderr, "Error: %s\n", cc ? cc->errstr : "OOM");
 }
 ```
 
+There are also several flags you can specify when using the `valkeyClusterOptions` helper struct.
```
zuiderkwast, line 75: Is it `.flags`? Is it a bitwise-or?

```suggestion
There are also several flags you can specify in `valkeyClusterOptions.flags`. It's a bitwise or of the following flags:
```

src/cluster.c
```diff
@@ -1301,52 +1359,32 @@ void valkeyClusterFree(valkeyClusterContext *cc) {
...
-    cc->flags = flags;
+valkeyClusterContext *valkeyClusterConnect(const char *addrs) {
+    valkeyClusterOptions options = {0};
+    options.initial_nodes = addrs;
 
-    return valkeyClusterConnectInternal(cc, addrs);
+    return valkeyClusterConnectWithOptions(&options);
 }
 
 valkeyClusterContext *valkeyClusterConnectWithTimeout(const char *addrs,
```
zuiderkwast, line 1381: Is this kept for symmetry with valkeyConnectWithOptions? And for easier migration?

src/cluster.c
```diff
@@ -2137,23 +2124,23 @@ static int prepareCommand(valkeyClusterContext *cc, struct cmd *command) {
...
-        cc->on_connect = fn;
+int valkeyClusterOptionsSetConnectCallback(valkeyClusterOptions *options,
+                                           void(fn)(const valkeyContext *c,
+                                                    int status)) {
+    if (options->connect_callback == NULL) {
+        options->connect_callback = fn;
         return VALKEY_OK;
     }
     return VALKEY_ERR;
 }
```
zuiderkwast, line 2135: Why is this kept? Can't the users just set this directly on the options object too?

src/cluster.c
```diff
@@ -2137,23 +2124,23 @@ static int prepareCommand(valkeyClusterContext *cc, struct cmd *command) {
...
-        cc->event_callback = fn;
-        cc->event_privdata = privdata;
+int valkeyClusterOptionsSetEventCallback(valkeyClusterOptions *options,
+                                         void(fn)(const valkeyClusterContext *cc,
+                                                  int event, void *privdata),
+                                         void *privdata) {
+    if (options->event_callback == NULL) {
+        options->event_callback = fn;
+        options->event_privdata = privdata;
         return VALKEY_OK;
```
zuiderkwast, line 2144: Why is this kept? Users can set this directly on the options object...?

src/cluster.c
```diff
@@ -2823,64 +2792,76 @@ valkeyClusterGetValkeyAsyncContext(valkeyClusterAsyncContext *acc,
     return ac;
 }
 
-valkeyClusterAsyncContext *valkeyClusterAsyncContextInit(void) {
+valkeyClusterAsyncContext *valkeyClusterAsyncContextInit(const valkeyClusterOptions *options) {
```
zuiderkwast, line 2795: Is there any point in supporting init without connect?

valkeyClusterAsyncConnectWithOptions would be enough, wouldn't it?

src/cluster.c
```diff
@@ -2823,64 +2792,76 @@ valkeyClusterGetValkeyAsyncContext(valkeyClusterAsyncContext *acc,
...
-                                            valkeyDisconnectCallback *fn) {
-    if (acc->onDisconnect == NULL) {
-        acc->onDisconnect = fn;
+int valkeyClusterAsyncSetEventCallback(valkeyClusterAsyncContext *acc,
+                                       void(fn)(const valkeyClusterContext *cc,
+                                                int event, void *privdata),
+                                       void *privdata) {
+    if (acc->cc->event_callback == NULL) {
+        acc->cc->event_callback = fn;
+        acc->cc->event_privdata = privdata;
```
zuiderkwast, line 2864: Can be done using the options object? Or not?

include/valkey/cluster.h
```diff
@@ -156,33 +144,49 @@ typedef struct valkeyClusterNodeIterator {
...
 } valkeyClusterNodeIterator;
 
-/*
- * Synchronous API
- */
+/* --- Configuration options --- */
+
+/* Enable slotmap updates using the command CLUSTER SLOTS.
+ * Default is the CLUSTER NODES command. */
+#define VALKEY_OPT_USE_CLUSTER_SLOTS 0x1000
```
zuiderkwast, line 151: Should we consider changing the default to SLOTS? It's recommended in Valkey.... But NODES is better for clusters with highly fragmented slot ranges.

bjosv: I like the idea of changing the default to SLOTS, specially if it's recommended.
We could add an option in the future that let the client decide and predicted which command is most efficient.

src/cluster.c
bjosv: `valkeyClusterAsyncContextInit` and `valkeyClusterAsyncSetEventCallback` are kept since we need to be able to use the created async-context when setting `privdata` in the event callback., i.e. chicken or the egg problem..
Example use-cases are when waiting for an VALKEYCLUSTER_EVENT_READY before sending the first async command in the event callback, using the async-context received in privdata.

I would be nice to get rid of them, but I have not yet found a solution for it.

src/cluster.c
bjosv: Lets remove this.

include/valkey/cluster.h
zuiderkwast: I think we should support both to make sure users can migrate without any changed characteristics or behaviour, but we can change the default. We can have a flag for VALKEY_OPT_USE_CLUSTER_NODES instead.

src/cluster.c
zuiderkwast: OK, let's keep it.

docs/cluster.md
zuiderkwast: Great. I hope you update the migration guide in parallel.

docs/cluster.md
```diff
@@ -194,18 +235,7 @@ The first alternative is to use the function `valkeyClusterAsyncConnect`, which
 Any command sent by the user thereafter will create a new non-blocking connection, unless a non-blocking connection already exists to the destination.
```
zuiderkwast, line 235: (I can't comment on the lines above.) The async API doc starts like this:

> ## Asynchronous API
> ### Connecting
> There are two alternative ways to initiate a cluster client, which also determines how the client behaves during the initial connect.
The first alternative is to use the function `valkeyClusterAsyncConnect`, which initially connects to the cluster in a blocking fashion and waits for the slot map before returning.

... and then the variant with init + connect2 is described.

But I checked `cluster.h` and I think we deleted the blocking connect and renamed connect2 to connect. We don't have any blocking valkeyClusterAsyncConnect anymore, do we?

I think the doc should first describe this variant:

1. Create options struct
2. Attach event loop to options
3. valkeyClusterAsyncConnectWithOptions

```C
valkeyClusterOptions options = {
    .initial_nodes = "127.0.0.1:7000";
};
valkeyClusterOptionsUseLibev(&options, EV_DEFAULT);
valkeyClusterAsyncContext *acc = valkeyClusterAsyncConnectWithOptions(&options);
```

and then the second variant with one more step `acc = init(options); connect(acc)`.

zuiderkwast (commented):
Looks mostly great. A large change that isn't very well explained in the docs is that you now attach the event loop to the options, instead of to the acc. We need to improve this part, also in the migration guide because it's an important and non-trivial change.

docs/migration-guide.md
```diff
@@ -27,14 +27,17 @@ The type `sds` is removed from the public API.
 
 ## Migrating from `hiredis-cluster` 0.14.0
 
-The command used to update the internal slot map is changed to `CLUSTER SLOTS`.
-`CLUSTER NODES` can be re-enabled through options, see `VALKEY_OPT_USE_CLUSTER_NODES`.
+* The cluster client initiation procedure is changed and `valkeyClusterOptions`
+  should be used to specify options when creating a context.
+  The `examples` directory in this repository contains some common client
+  initiation examples that might be helpful.
```
zuiderkwast, line 33: We can also link to cluster.md from this migration guide, right? I think cluster.md should have some example of async connect with options.

docs/cluster.md
```diff
@@ -27,37 +29,61 @@ It is not intended as a complete reference. For that it's always best to refer t
...
 ```c
-valkeyClusterContext *cc = valkeyClusterContextInit();
-valkeyClusterSetOptionAddNodes(cc, "127.0.0.1:6379,127.0.0.1:6380");
-int status = valkeyClusterConnect2(cc);
-if (status == VALKEY_ERR) {
-    printf("Error: %s\n", cc->errstr);
-    // handle error
+valkeyClusterOptions opt = {0};
+opt.initial_nodes = "127.0.0.1:6379,127.0.0.1:6380"; // Addresses to initially connect to.
+opt.options = VALKEY_OPT_USE_CLUSTER_NODES;          // See available flags below.
```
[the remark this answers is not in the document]
bjosv: I used the name `options` for this field to match how it's named in `valkeyOptions` (which also has an `options` which takes `VALKEY_OPT_*`.).  I guess we should change both in that case.

When contexts are initiated we keep the given option-flags in a context-field named flags, but here we also use it to keep internal state, both in cluster contexts and standalone.

docs/cluster.md
```diff
@@ -27,37 +29,61 @@ It is not intended as a complete reference. For that it's always best to refer t
...
+if (cc == NULL || cc->err) {
+    fprintf(stderr, "Error: %s\n", cc ? cc->errstr : "OOM");
 }
 ```
 
+There are also several flags you can specify in `valkeyClusterOptions.flags`. It's a bitwise OR of the following flags:
+
+| Flag | Description  |
+| --- | --- |
+| VALKEY\_OPT\_USE\_CLUSTER\_NODES | Tells libvalkey to use the command `CLUSTER NODES` when updating its slot map (cluster topology).<br>Libvalkey uses `CLUSTER SLOTS` by default. |
```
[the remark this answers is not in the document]
bjosv: ah nice! That makes it readable when "raw-reading" from a terminal, will update standalone.md too.

docs/cluster.md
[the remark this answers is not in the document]
zuiderkwast: We can keep options then. If we change to flags, it would imply all the macro names that contain VALKEY_OPT too. It would be too big change maybe.

docs/cluster.md
bjosv: Oh, this part doesn't seem finished, there are a lot of empty lines too :(  I will update.

docs/cluster.md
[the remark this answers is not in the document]
bjosv: Yes, it getting large then. We can think about it, and possibly replace them all in a separate PR afterwards.

docs/cluster.md
```diff
@@ -227,29 +225,64 @@ The `err` field in the `valkeyContext` can be used to find out the cause of the
...
+// Set the event callback using the context as privdata.
+valkeyClusterAsyncSetEventCallback(acc, eventCallback, acc);
+
+// Start connecting to the initial nodes.
+valkeyClusterAsyncConnect(acc)
+```
 
 ### Connection options
 
+There are a variety of options you can specify using the `valkeyClusterOptions` struct when connecting to a cluster.
```
zuiderkwast, line 280:
```suggestion
There is a variety of options you can specify using the `valkeyClusterOptions` struct when connecting to a cluster.
```

zuiderkwast (commented):
Great docs!

Ready to merge, or are you considering some more update to the migration guide?

src/cluster.c
bjosv: Yes, it was kept for symmetry.
If we change the default timeouts to something reasonable the `valkeyClusterConnectWithTimeout` API is less useful and the `valkeyClusterConnect` API could easily be used instead when migrating.

zuiderkwast: OK, we can still keep it for symmetry AND change the default timeout I guess.

docs/migration-guide.md
bjosv: Added links in the migration guide to the newly added examples in cluster.md

docs/cluster.md
bjosv: Corrected

docs/migration-guide.md
```diff
@@ -27,11 +27,20 @@ The type `sds` is removed from the public API.
 
 ## Migrating from `hiredis-cluster` 0.14.0
 
+* The cluster client initiation procedure is changed and `valkeyClusterOptions`
+  should be used to specify options when creating a context.
+  See documentation for configuration examples when using the
+  [Synchronous API](cluster.md#connection-options) or the
+  [Asynchronous API](cluster.md#connecting-1).
```
zuiderkwast, line 34: Why not link to the headings mentioned in the link text?

How to connect is the first subsection within these sections anyway.

```suggestion
  [Synchronous API](cluster.md#synchronous-api) or the
  [Asynchronous API](cluster.md#asynchronous-api).
```

zuiderkwast (approved):

docs/migration-guide.md
bjosv: Sure

PR #140, opened 2024-12-13
Rename offensive defines and members in API

bjosv:
- Rename offensive cluster role defines in the API:
  Use `VALKEY_ROLE_PRIMARY` instead of `VALKEY_ROLE_MASTER`.
  Use `VALKEY_ROLE_REPLICA` instead of `VALKEY_ROLE_SLAVE`.
  Use `VALKEY_ROLE_UNKNOWN` instead of `VALKEY_ROLE_NULL`.
- Rename `valkeyClusterNode` member `slaves` to `replicas`.
- Rename offensive variables.
- Improve wording in migration guide ([Review comment](https://github.com/valkey-io/libvalkey/pull/122#discussion_r1858760013)).


zuiderkwast (approved):
LGTM

When I saw "use UNKNOWN instead of NULL" in the PR description, I first thought it's about renaming the NULL pointers. (I though what!? Is NULL an offesive word?), but then I realized it's about the unknown role. So maybe formulate this in another way, for example use the full macro names instead of a bare `NULL`. :)

mkmkme (approved):

PR #142, opened 2024-12-18
Simplify API by only using non-const connect callbacks

bjosv:
Remove `valkeyAsyncSetConnectCallbackNC` since it was first added in hiredis for backwards-compatibility reasons,
and let `valkeyAsyncSetConnectCallback` use the non-const connect callback prototype.

Includes the removal of the cluster version `valkeyClusterAsyncSetConnectCallbackNC` and updates of tests.

### DIFF
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -15,11 +15,15 @@ The general actions needed are:
 
 The type `sds` is removed from the public API.
 
+### Renamed API functions
+
+* `redisAsyncSetConnectCallbackNC` is renamed to `valkeyAsyncSetConnectCallback`.
+
 ### Removed API functions
 
 * `redisFormatSdsCommandArgv` removed from API. Can be replaced with `valkeyFormatCommandArgv`.
 * `redisFreeSdsCommand` removed since the `sds` type is for internal use only.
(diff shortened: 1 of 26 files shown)


docs/migration-guide.md
```diff
@@ -19,6 +19,7 @@ The type `sds` is removed from the public API.
 
 * `redisFormatSdsCommandArgv` removed from API. Can be replaced with `valkeyFormatCommandArgv`.
 * `redisFreeSdsCommand` removed since the `sds` type is for internal use only.
+* `redisAsyncSetConnectCallbackNC` removed and `valkeyAsyncSetConnectCallback` accepts the non-const callback function prototype.
```
zuiderkwast, line 22: Shouldn't we also mention `redisAsyncSetConnectCallback` on it's own line?

It's renamed in common way to `valkeyAsyncSetConnectCallback`, but the const modifier is removed.

bjosv:
Well spotted, seems like the spellchecker in CI work quite fine as well.

zuiderkwast (approved):
Ping @michael-grunder

michael-grunder (approved):
LGTM. The original weirdness was a bit of a hack.

PR #148, opened 2025-01-15
Handle empty addresses in `CLUSTER NODES` responses

bjosv:
When parsing a `CLUSTER NODES` response we will now treat an empty IP string as it means the same endpoint as the response came from. This due to that the docs mention that an empty string is returned for the IP field when the node doesn't know its own IP address.
The created `valkeyClusterNode` will get its address from the `valkeyContext` used when sending the command.

Additional changes are that we will use the `noaddr` flag instead of the address length when deciding when a node should be skipped; and added validation of the port.

Example of a response snippet with an empty address:
"752d150249c157c7cb312b6b056517bbbecb42d2 **:6379**@&#65279;16379 myself,master - 1658754833817 1658754833000 3 connected 5461-10922"

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -753,7 +753,7 @@ static int store_replica_nodes(dict *nodes, dict *replicas) {
  * Only parse primary nodes if the `parsed_primary_id` argument is NULL,
  * otherwise replicas are also parsed and its primary_id is returned by pointer
  * via 'parsed_primary_id'. */
-static int parse_cluster_nodes_line(valkeyClusterContext *cc, char *line,
+static int parse_cluster_nodes_line(valkeyClusterContext *cc, valkeyContext *c, char *line,
                                     valkeyClusterNode **parsed_node, char **parsed_primary_id) {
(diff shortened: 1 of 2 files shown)


src/cluster.c
```diff
@@ -753,7 +753,7 @@ static int store_replica_nodes(dict *nodes, dict *replicas) {
  * Only parse primary nodes if the `parsed_primary_id` argument is NULL,
  * otherwise replicas are also parsed and its primary_id is returned by pointer
  * via 'parsed_primary_id'. */
-static int parse_cluster_nodes_line(valkeyClusterContext *cc, char *line,
+static int parse_cluster_nodes_line(valkeyClusterContext *cc, valkeyContext *c, char *line,
```
zuiderkwast, line 756: Mention what `c` is in the comment above.

The return value VALKEY_OK/ERR isn't mentioned either.

zuiderkwast (approved):
Awesome.

We could mention or quote something from the docs that motivate why we do this, in the commit message / PR top comment.

src/cluster.c
zuiderkwast: Updated comment looks good.

PR #149, opened 2025-01-17
Add cluster slots parsing tests

bjosv:
Adds `CLUSTER SLOTS` parser tests, which uses a testsuite-local primitive parser of a JSON subset to encode arrays to RESP.

### DIFF
--- a/tests/ut_slotmap_update.c
+++ b/tests/ut_slotmap_update.c
@@ -19,20 +19,106 @@ const char *__asan_default_options(void) {
 
 #include <stdbool.h>
 
+valkeyReply *create_reply(const char *buf, size_t len);
+char *resp_encode_array(char *p, sds *resp);
+
 /* Helper to create a valkeyReply that contains a bulkstring. */
-valkeyReply *create_cluster_nodes_reply(const char *bulkstr) {
-    valkeyReply *reply;
-
+valkeyReply *create_cluster_nodes_reply(const char *str) {
     /* Create a RESP Bulk String. */
-    char cmd[1024];
-    int len = sprintf(cmd, "$%zu\r\n%s\r\n", strlen(bulkstr), bulkstr);
+    char buf[1024];
+    int len = sprintf(buf, "$%zu\r\n%s\r\n", strlen(str), str);
 
-    /* Create a valkeyReply. */
+    return create_reply(buf, len);
+}
+
+/* Helper to create a cluster slots response.
+ * Parses the string using a rudimentary JSON like format which accepts:
+ * - arrays   example: [elem1, elem2]
+ * - strings  example: 'mystring'
+ * - integers example: 123
+ * - null     example: null
+ * See resp_encode_array for details.
+ */
+valkeyReply *create_cluster_slots_reply(const char *str) {
+    sds resp = sdsempty();
+
+    char *s = strdup(str);
+    resp_encode_array(s, &resp);
+    free(s);
+
+    valkeyReply *reply = create_reply(resp, sdslen(resp));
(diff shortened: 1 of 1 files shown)


tests/ut_slotmap_update.c
```diff
@@ -426,6 +512,181 @@ void test_parse_cluster_nodes_with_legacy_format(void) {
...
+    cluster_slot *slot;
+    dictIterator di;
+
+    if (parse_replicas)
+        cc->flags |= VALKEYCLUSTER_FLAG_ADD_SLAVE;
+
+    valkeyReply *reply = create_cluster_slots_reply(
+        "[[0, 5460, ['127.0.0.1', 30001, 'e7d1eecce10fd6bb5eb35b9f99a514335d9ba9ca', ['hostname', 'localhost']],"
+        "           ['127.0.0.1', 30004, '07c37dfeb235213a872192d90877d0cd55635b91', ['hostname', 'localhost']]],"
+        " [5461, 10922, ['127.0.0.1', 30002, '67ed2db8d677e59ec4a4cefb06858cf2a1a89fa1', ['hostname', 'localhost']],"
```
zuiderkwast, line 528: Interesting! I guess it's pretty easy to convert this pseudo-JSON to RESP, althought it adds some non-trivial code to the testing. My first though is that single-quoted stings are not allowed in JSON. :)

My second thought is that it might be possible to construct a parse tree structure that is readable and easy to serialize to RESP, without parsing. Test code should be trivial, so there is no doubt if there's a bug, is it a bug in the test code or in the real implementation.

Here is a reply literal with the first slot range (untested). It's much bigger than the pseudo-JSON obviously.

```C
valkeyReply reply = {.type = VALKEY_REPLY_ARRAY, .elements = 3, .element = (valkeyReply[]){
    {.type = VALKEY_REPLY_ARRAY, .elements = 4, .element = (valkeyReply[]){
        {.type = VALKEY_REPLY_INTEGER, .integer = 0},
        {.type = VALKEY_REPLY_INTEGER, .integer = 5460},
        {.type = VALKEY_REPLY_ARRAY, .elements = 4, .element = (valkeyReply[]){
            {.type = VALKEY_REPLY_STRING, .str = "127.0.0.1"},
            {.type = VALKEY_REPLY_INTEGER, .integer = 30005},
            {.type = VALKEY_REPLY_STRING, .str = "e7d1eecce10fd6bb5eb35b9f99a514335d9ba9ca"},
            {.type = VALKEY_REPLY_ARRAY, .elements = 2, .element = (valkeyReply[]){
                {.type = VALKEY_REPLY_STRING, .str = "hostname"},
                {.type = VALKEY_REPLY_STRING, .str = "localhost"} 
            }
        }
    },
    ...
};
```

bjosv: Interesting, are there similar tests in the valkey repo?

zuiderkwast: No, but similar nested struct literals, in some modules used for module API tests. The one for testing key specifications IIRC.

zuiderkwast: The tests for replies like CLUSTER SLOTS use Tcl code which is compact but most often doesn't check the types.

A nested reply like that is a Tcl list on the form

    {{0 5460 {127.0.0.1 30001 abcde123456 {hostname localhost} ...

zuiderkwast: Let's use the pseudo-JSON then. It's pretty simple.

bjosv: I'll merge this variant then.

PR #159, opened 2025-01-21
Retry when an async slotmap update fails 

bjosv:
Small fix to reattempt failing slotmap updates.

Additionally:

- Handle empty and NULL addresses in `CLUSTER SLOTS` responses.  
    As described in cluster slots docs:
  `Clients may treat the empty string in the same way as NULL,
     that is the same endpoint it used to send the current command to`
    
 - `clusterclient_async` received a new flag `--async-initial-update` to enable usage of `valkeyClusterAsyncConnect2`, which avoids blocking initial slotmap update.
  - Use short timeout when scheduling processing of next command in `clusterclient_async` and allow the event engine to read from sockets before next command, which improves predictability in tests.
     
Final change similar to https://github.com/Nordix/hiredis-cluster/pull/252.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -389,37 +389,9 @@ error:
  * Return a new node with the "cluster slots" command reply.
  */
 static valkeyClusterNode *node_get_with_slots(valkeyClusterContext *cc,
-                                              valkeyReply *host_elem,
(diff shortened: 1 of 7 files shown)


zuiderkwast (approved):
LGTM

PR #165, opened 2025-02-03
Remove APIs for separate init and connect of async cluster contexts

bjosv:
Embed a `valkeyClusterContext` in a created `valkeyClusterAsyncContext` to be able to cast between these types. This enables users to access the async context in an event callback without using `privdata`, and we can remove the APIs previously kept just for this scenario:
```
valkeyClusterAsyncContext *valkeyClusterAsyncContextInit(const valkeyClusterOptions *options);
int valkeyClusterAsyncConnect(valkeyClusterAsyncContext *acc);
int valkeyClusterAsyncSetEventCallback(valkeyClusterAsyncContext *acc,
                                       void(fn)(const valkeyClusterContext *cc,
                                                int event, void *privdata),
```

Also removing the testcase based on `clusterclient_reconnect_async.c` since it was using the removed APIs.

There is now a symmetry to how standalone contexts works, i.e. `valkeyAsyncContext` embeds the `valkeyContext`.


src/cluster.c
```diff
@@ -2765,13 +2757,21 @@ valkeyClusterAsyncContext *valkeyClusterAsyncContextInit(const valkeyClusterOpti
         return NULL;
     }
 
-    acc = vk_calloc(1, sizeof(valkeyClusterAsyncContext));
+    acc = vk_realloc(cc, sizeof(valkeyClusterAsyncContext));
```
zuiderkwast, line 2760: We could allocate once. It may be even simpler.

Before we get here, we do `cc = valkeyClusterContextInit(options);`. `valkeyClusterContextInit` is internal so we can change it. We can remove the allocation from it and change it to a void function and we can pass the allocation to it, like this:

```
    acc = vk_calloc(1, sizeof(valkeyClusterAsyncContext));
    valkeyClusterContextInit((valkeyClusterContext *)acc, options);
    if (acc->cc.err != 0) {
        ...
    }
```

WDYT?

include/valkey/cluster.h
```diff
@@ -289,17 +292,6 @@ valkeyClusterAsyncContext *valkeyClusterAsyncConnectWithOptions(const valkeyClus
 void valkeyClusterAsyncDisconnect(valkeyClusterAsyncContext *acc);
 void valkeyClusterAsyncFree(valkeyClusterAsyncContext *acc);
 
-/* Initiate and connect as separate steps. */
-valkeyClusterAsyncContext *valkeyClusterAsyncContextInit(const valkeyClusterOptions *options);
-int valkeyClusterAsyncConnect(valkeyClusterAsyncContext *acc);
-
-/* Callback option configurable after a context initiation, enabling that the
- * valkeyClusterAsyncContext pointer can be given as privdata in the callback. */
-int valkeyClusterAsyncSetEventCallback(valkeyClusterAsyncContext *acc,
```
zuiderkwast, line 298: Mention in the migration guide that these are deleted?

docs/cluster.md
```diff
@@ -339,9 +320,14 @@ After this, the disconnection callback is executed with the `VALKEY_OK` status a
 
 Use [`event_callback` in `valkeyClusterOptions`](#events-per-cluster-context) to get notified when certain events occur.
 
-When the callback function requires the current `valkeyClusterAsyncContext` it can be provided in the `privdata`.
-In this case initiate the context using `valkeyClusterAsyncContextInit`, set the callback and `privdata` using `valkeyClusterAsyncSetEventCallback`,
-and initiate connection attempts using `valkeyClusterAsyncConnect` as described under the [Connecting](#connecting-1) section.
+When the callback function requires the current `valkeyClusterAsyncContext` typecast the given `valkeyClusterContext` to a `valkeyClusterAsyncContext`.
+A `valkeyClusterAsyncContext` struct begins with the embedded `valkeyClusterContext` so these structs share memory address.
```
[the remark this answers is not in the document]
bjosv: I'll use your suggestion since its shorter and sounds good.
This text is only under the `Asynchronous API` section in the markdown, so hopefully it should be clear when reading the whole section.

src/cluster.c
bjosv: Thanks, I believe it got better changing this.

include/valkey/cluster.h
bjosv: Added to the removed section in migration guide.

PR #166, opened 2025-02-05
SSL_library_init() is deprecated as of openssl v1.1.0

michael-grunder:
According to the openss docs `SSL_library_init()` is deprecated as of openssl v1.1.0. We still need to call it when using crypto locks because we define that if openssl is < 1.1.0.

See: https://manpages.debian.org/experimental/libssl-doc/SSL_library_init.3ssl.en.html
See: https://docs.openssl.org/3.3/man3/OPENSSL_init_ssl/


bjosv (approved):
A related issue: https://github.com/valkey-io/valkey/issues/1192

PR #169, opened 2025-02-17
Rework library naming on `macOS`.

michael-grunder:
The idiom on macOS is to put the version before the `dylib` extension instead of after it like Linux.


bjosv (approved):
That's different. I this handled correctly by CMake out the box?

michael-grunder:
It looks like it, yeah.

```bash
-rwxr-xr-x  1 michaelgrunder  staff  211776 Feb 18 12:33 libvalkey.0.1.dylib
lrwxr-xr-x  1 michaelgrunder  staff      19 Feb 18 12:33 libvalkey.dylib -> libvalkey.0.1.dylib
```

PR #173, opened 2025-03-04
Refactor `dict`

bjosv:
To be able to use `dict` from Valkey I have found that following changes would be needed

- Rename `dictSetHash*` to `dictSet*` (matching API)
- Rename `dictGetEntry*` to `dictGet*`  (matching API)
- Make `dictEntry` opaque    
    This matches the `dictEntry` is valkey which also is opaque.
- Remove unused `privdata` from dict.    
    Since the `privdata` pointer is not used in libvalkey we can remove it.
- Initiate dictType's using designated initializers (C99)
Allows us to handle different ordering in the dictType struct.
 - Duplicate dict value explicitly to avoid using `valDup`   
   Valkeys dict type don't have `valDup`, i.e a function that can copy the inserted data.
   We can avoid this difference by manually copy the value.
- Update dictType signatures to match dict in Valkey.
  `callbackHash` returns an unsigned long int in valkey, we can use that as well.   

With these changes we would be able to build libvalkey using `dict` from valkeys repos in its *current state*.


src/cluster.c
```diff
@@ -120,13 +120,12 @@ void listClusterNodeDestructor(void *val) { freeValkeyClusterNode(val); }
 
 void listClusterSlotDestructor(void *val) { cluster_slot_destroy(val); }
 
-unsigned int dictSdsHash(const void *key) {
+unsigned long int dictSdsHash(const void *key) {
```
zuiderkwast, line 123: Is there a risk this and other functions are defined in valkey too with a different hash function under the hood? Any risk we or valkey will use the wrong one if both are statically linked in the same binary, or will we get an link-time error if that happens?

zuiderkwast (approved):
Looks like a cleanup. More lines deleted than added. That's always nice.

PR #174, opened 2025-03-04
Update CI

bjosv:
- Updates the apt package index before installing valkey in CI.
- Replace `ubuntu-20.04` since the runner will be removed 2025-04-01.
- Change gcc/clang to use runner available versions.
- Update use of latest clang to clang-19.


bjosv (commented):
Approved in  PR 173

PR #179, opened 2025-03-13
Fix Coverity warnings

bjosv:
- Remove inconsistent check of `c->funcs` in `valkeyReconnect`.  
    Make sure we always set a `valkeyContextFuncs` before returning from `valkeyConnectWithOptions()`.
  We will then always have a `c->funcs` when calling `c->funcs->close`, `c->funcs->connect` and more in multiple places.
    
    **Fixes**: CID443767 "Comparing c->funcs to null implies that c->funcs might be null"

- Fix coverity issues in test.
   Check context before member of the context.
   
   **Fixes:** CID 456643: Dereference null return value (NULL_RETURNS)

    Use memcpy instead of strcpy to correct:
    "You might overrun the 256-character fixed-size string astest.errstr by copying c->errstr without checking the length."
    
    **Fixes:**
CID 443771: Copy into fixed size buffer (STRING_OVERFLOW)
                       CID 443768: Copy into fixed size buffer (STRING_OVERFLOW)
                       CID 443753: Copy into fixed size buffer (STRING_OVERFLOW)

### DIFF
--- a/src/valkey.c
+++ b/src/valkey.c
@@ -773,10 +773,10 @@ int valkeyReconnect(valkeyContext *c) {
     memset(c->errstr, '\0', strlen(c->errstr));
 
     assert(c->funcs);
-    if (c->funcs->close)
+    if (c->funcs && c->funcs->close)
         c->funcs->close(c);
 
-    if (c->privctx && c->funcs->free_privctx) {
+    if (c->privctx && c->funcs && c->funcs->free_privctx) {
         c->funcs->free_privctx(c->privctx);
         c->privctx = NULL;
     }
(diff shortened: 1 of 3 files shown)


src/valkey.c
```diff
@@ -772,9 +772,9 @@ int valkeyReconnect(valkeyContext *c) {
     c->err = 0;
     memset(c->errstr, '\0', strlen(c->errstr));
 
-    if (c->funcs && c->funcs->close) {
+    assert(c->funcs);
```
michael-grunder, line 775: Maybe it's acceptable but I think this will still lead to UB if libvalkey is compiled with `-DNDEBUG`.

michael-grunder (changes requested):
LGTM other than the `-DNDEBUG` thing if we care :smile:

Great hardening though.

src/valkey.c
bjosv: Ah.. `valkeyContextSetFuncs()` should set c->funcs, but its assert is not checked then either I guess..
I'll re-add the check here, and add checks in all other c->funcs calls in valkeyReconnect.

michael-grunder (approved):

PR #182, opened 2025-03-21
Updates needed for replacing hiredis in the valkey repo

bjosv:
- Add `SDS_INCLUDE_DIR` and `DICT_INCLUDE_DIR` to be able to replace sds and dict.
- Handle `sdsrange()` API difference
The  `sdsrange` in Valkey return void and performs no length check.
- Use same return type as Valkey from the hashFunction in dictType
Fixes issues on 32bit and macOS
- Update the nodeIterator memory blob to fit Valkeys larger dict iterator.

Changes needed in valkey repo:
https://github.com/valkey-io/valkey/compare/unstable...bjosv:valkey:refs/heads/use-libvalkey

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -13,6 +13,8 @@ IncludeCategories:
     Priority: -1
   - Regex: '^"'
     Priority: 1
+  - Regex: '^<(sds|dict).h>'
+    Priority: 2
   - Regex: '^<valkey/adapters/'
     Priority: 3
   - Regex: '^<valkey/'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,7 @@ else()
   add_definitions(-D_CRT_SECURE_NO_WARNINGS -DWIN32_LEAN_AND_MEAN)
 endif()
 
-SET(valkey_sources
+set(valkey_sources
     src/adlist.c
     src/alloc.c
     src/async.c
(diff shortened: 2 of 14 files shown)


zuiderkwast:
> * Moved `sds` and `dict` to own directory.
> 
> * Add Makefile option `USE_INCLUDE` to be able to replace sds and dict with types provided in the Valkey repo.
>   The build in valkey will use `USE_INCLUDE=../../src/` (relative to deps) to provide its `sds.h` and `dict.h`.

Do they need to be in a separate sub directory? Can't they stay where they are and when it's built and linked by valkey, we just give precedence to valkey's own src dir?

I could also accept specific variables like SDS_INCLUDE_DIR, DICT_INCLUDE_DIR, etc. to explicitly indicate these two libs.

In your valkey branch, I don't see how you link libvalkey with valkey's sds and dict. When linking e.g. valkey-cli, we link to `libvalkey.a`, which is built without dict and sds?

bjosv:
> Do they need to be in a separate sub directory? Can't they stay where they are and when it's built and linked by valkey, we just give precedence to valkey's own src dir?

It should be possible, this was to follow the initial talks how it could be done. I'll try it so we can see how it looks.

> In your valkey branch, I don't see how you link libvalkey with valkey's sds and dict. When linking e.g. valkey-cli, we link to `libvalkey.a`, which is built without dict and sds?

When building valkey-cli I added `sds.o` and sds's dependencies to ENGINE_CLI_OBJ.
When linking valkey-cli with libvalkey the linker already knows sds and since the cli is not using, nor will incorporate, the `cluster.c` or `async.c` code the `dict` type is not required.

`ENGINE_CLI_OBJ=anet.o ... sds.o util.o sha256.o`

zuiderkwast (approved):

zuiderkwast:
Still draft?

bjosv:
> Still draft?

Seems ready now. The test-build to use the latest changes passed in valkey.

zuiderkwast:
OK, feel free to merge then.

PR #184, opened 2025-03-25
Add CI job that builds with RDMA support enabled

bjosv:
The job will make sure that `ENABLE_RDMA` / `USE_RDMA` flags builds.

Includes 2 corrections:
 - Makefile: Let the RDMA library be built and installed using `make install`.
   Updated `make install` to depend on the `install-rdma` target when RDMA is enabled.
    Includes some additional cleanups.
 - CMake: Correcting the link order when building `client_test`.    
    When building with CMake using ENABLE_RDMA=ON and BUILD_SHARED_LIBS=OFF the build fails.
   `libvalkey_rdma.a` uses `vk_usec_now` from `libvalkey.a`, and this is fixed by correcting the link order.


bjosv:
@pizhenwei I was not able to set you as a reviewer, but please check so I haven't destroyed anything you need.

pizhenwei:
> @pizhenwei I was not able to set you as a reviewer, but please check so I haven't destroyed anything you need.

LGTM.

PR #189, opened 2025-04-08
Introduce MPTCP

pizhenwei:
Multipath TCP (MPTCP) is an extension of the standard TCP protocol that allows a single transport connection to use multiple network interfaces or paths. MPTCP is useful for applications like bandwidth aggregation, failover, and more resilient connections.

Linux kernel starts to support MPTCP since v5.6, it's time to support it.

### DIFF
--- a/include/valkey/valkey.h
+++ b/include/valkey/valkey.h
@@ -100,6 +100,9 @@ typedef SSIZE_T ssize_t;
 #define VALKEY_PREFER_IPV4 0x800
 #define VALKEY_PREFER_IPV6 0x1000
 
+/* Flag specific to use Multipath TCP (MPTCP) */
+#define VALKEY_MPTCP 0x2000
+
 #define VALKEY_KEEPALIVE_INTERVAL 15 /* seconds */
 
 /* number of times we retry to connect in the case of EADDRNOTAVAIL and
@@ -169,7 +172,8 @@ enum valkeyConnectionType {
 #define VALKEY_OPT_PREFER_IPV4 0x20       /* Prefer IPv4 in DNS lookups. */
 #define VALKEY_OPT_PREFER_IPV6 0x40       /* Prefer IPv6 in DNS lookups. */
 #define VALKEY_OPT_PREFER_IP_UNSPEC (VALKEY_OPT_PREFER_IPV4 | VALKEY_OPT_PREFER_IPV6)
-#define VALKEY_OPT_LAST_SA_OPTION 0x40 /* Last defined standalone option. */
+#define VALKEY_OPT_MPTCP 0x80
+#define VALKEY_OPT_LAST_SA_OPTION 0x80 /* Last defined standalone option. */
 
 /* In Unix systems a file descriptor is a regular signed int, with -1
  * representing an invalid descriptor. In Windows it is a SOCKET
@@ -232,6 +236,14 @@ typedef struct {
         (opts)->type = VALKEY_CONN_TCP;          \
         (opts)->endpoint.tcp.ip = ip_;           \
         (opts)->endpoint.tcp.port = port_;       \
+    } while (0)
+
+#define VALKEY_OPTIONS_SET_MPTCP(opts, ip_, port_) \
+    do {                                           \
+        (opts)->type = VALKEY_CONN_TCP;            \
+        (opts)->endpoint.tcp.ip = ip_;             \
+        (opts)->endpoint.tcp.port = port_;         \
+        (opts)->options = VALKEY_OPT_MPTCP;        \
     } while (0)
 
 #define VALKEY_OPTIONS_SET_UNIX(opts, path)  \
--- a/include/valkey/valkey.h
+++ b/include/valkey/valkey.h
@@ -243,7 +243,7 @@ typedef struct {
         (opts)->type = VALKEY_CONN_TCP;            \
         (opts)->endpoint.tcp.ip = ip_;             \
         (opts)->endpoint.tcp.port = port_;         \
-        (opts)->options = VALKEY_OPT_MPTCP;        \
+        (opts)->options |= VALKEY_OPT_MPTCP;       \
     } while (0)
 
 #define VALKEY_OPTIONS_SET_UNIX(opts, path)  \
--- a/src/net.c
+++ b/src/net.c
@@ -43,6 +43,7 @@
 
 #include <sds.h>
 
+#include <assert.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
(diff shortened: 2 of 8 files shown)


pizhenwei:
Hi,

The valkey server [MPTCP support PR](https://github.com/valkey-io/valkey/pull/1811) is not merged yet, please wait for a few days.

src/net.c
```diff
@@ -374,6 +374,25 @@ int valkeyTcpSetTimeout(valkeyContext *c, const struct timeval tv) {
...
+        (void)(c);
+        return IPPROTO_MPTCP;
+    }
+#else
+        valkeySetErrorFromErrno(c, VALKEY_ERR_PROTOCOL, "MPTCP is not supported on this platform");
+        return VALKEY_ERR;
+    }
+#endif
+    return ai_protocol;
+}
```
zuiderkwast, line 394: Same discussion as we had in the Valkey PR https://github.com/valkey-io/valkey/pull/1811

Let's change this in a similar way?

```suggestion
static int valkeyContextTcpGetProtocol(int is_mptcp_enabled) {
#ifdef IPPROTO_MPTCP
    return is_mptcp_enabled ? IPPROTO_MPTCP : IPPROTO_TCP;
#else
    return IPPROTO_TCP;
#endif
}
```

src/net.c
```diff
@@ -454,7 +473,11 @@ int valkeyContextConnectTcp(valkeyContext *c, const valkeyOptions *options) {
     }
     for (p = servinfo; p != NULL; p = p->ai_next) {
     addrretry:
-        if ((s = socket(p->ai_family, p->ai_socktype, p->ai_protocol)) == VALKEY_INVALID_FD)
+        rv = valkeyContextSetMptcp(c, p->ai_protocol);
+        if (rv == VALKEY_ERR)
+            goto error;
+
+        if ((s = socket(p->ai_family, p->ai_socktype, rv)) == VALKEY_INVALID_FD)
```
zuiderkwast, line 480: Let's change this in the same way we changed it in the Valkey PR?

```suggestion
        rv = valkeyContextTcpGetProtocol(c->flags & VALKEY_MPTCP);
        if ((s = socket(p->ai_family, p->ai_socktype, rv)) == VALKEY_INVALID_FD)
```

src/valkey.c
```diff
@@ -854,6 +854,10 @@ valkeyContext *valkeyConnectWithOptions(const valkeyOptions *options) {
         c->flags |= VALKEY_PREFER_IPV6;
     }
 
+    if (options->options & VALKEY_OPT_MPTCP) {
+        c->flags |= VALKEY_MPTCP;
+    }
```
zuiderkwast, line 859: Shall we set the error here?

```suggestion
    if (options->options & VALKEY_OPT_MPTCP) {
#ifndef IPPROTO_MPTCP
        valkeySetError(c, VALKEY_ERR_PROTOCOL, "MPTCP is not supported on this platform");
        return c;
#endif
        c->flags |= VALKEY_MPTCP;
    }
```

zuiderkwast (commented):
Nice!

If the server doesn't support MPTCP, it will silently fallback to plain TCP, so the tests will pass on all versions of Valkey?

I have some comments very similar to the comments I had on the Valkey PR.

src/net.c
pizhenwei: There is a config check for `IPPROTO_MPTCP` in valkey, so `valkeyContextTcpGetProtocol` will not fail.
sadly we don't have the same check in libvalkey.

src/valkey.c
zuiderkwast: This file doesn't have IPPROTO_MPTCP because it doesn't `#include` the header where it is defined. Maybe this is not the right place to do the feature check...

When the async API is used, the connect and creation of the context are separated. Connection is done like this:

1. `valkeyAsyncConnectWithOptions()`
2. Attach event loop to async context
3. Run even loop

The context is created in step 1 but the actual connection happens in step 3. I think it's better if we can get the error in step 1. What do you think?

bjosv: I believe the call to connect is done in step 1, as a non-blocking connect. Then in step 3 the socket events will be handled, i.e the client will see that the connect happens.

zuiderkwast: Ah, so in that case, it's OK to do the feature check in `valkeyConnectWithOptions()`.

zuiderkwast: Is this true also for the cluster counterpart?

pizhenwei:
Hi @zuiderkwast @bjosv 
Please take a look at the latest version, thanks!

src/net.c
```diff
@@ -374,6 +375,32 @@ int valkeyTcpSetTimeout(valkeyContext *c, const struct timeval tv) {
...
+
+/* XXX: Until glibc 2.41, getaddrinfo with hints.ai_protocol of IPPROTO_MPTCP leads error.
+ * Use hints.ai_protocol IPPROTO_IP (0) or IPPROTO_TCP (6) to resolve address and overwrite
+ * it when MPTCP is enabled.
+ * Ref: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/net/mptcp/mptcp_connect.c
+ *      https://sourceware.org/git/?p=glibc.git;a=commit;h=a8e9022e0f829d44a818c642fc85b3bfbd26a514
+ */
+static int valkeyTcpGetProtocol(int is_mptcp_enabled) {
+    return is_mptcp_enabled ? IPPROTO_MPTCP : IPPROTO_TCP;
+}
```
bjosv, line 391: I think it would be easier to read if we ifdef within the functions, WDYT?
```suggestion
static int valkeyTcpGetProtocol(int is_mptcp_enabled) {
#ifdef IPPROTO_MPTCP
    return is_mptcp_enabled ? IPPROTO_MPTCP : IPPROTO_TCP;
#else
    assert(!is_mptcp_enabled);
    return IPPROTO_TCP;
#endif /* IPPROTO_MPTCP */
}
```

Same comment for `valkeyHasMptcp`.

pizhenwei: This is the previous version, I converted to current version. Hahaha~
I'm not sure which style is better...

pizhenwei: Someday if `IPPROTO_MPTCP` is popular enough, just removing the `#else` branch is OK...

zuiderkwast: I think it doesn't matter. Both are OK. I would leave it up to the author. :)

But we need to keep `#else` for a long time, because we support many platforms and some of them will probably never have MPTCP within a near future.

bjosv (approved):
LGTM!

src/net.c
bjosv: This was just a minor nit to have a similar look as its (currently) done in valkey. Current is fine.

pizhenwei:
Hi @michael-grunder 
MPTCP support is merged into server side, would you please take a look at this PR?

michael-grunder (approved):
LGTM

zuiderkwast (commented):
I think we can also mention it in the documentation.

* README under Features (somewhere near where we mention RDMA)
* Documentation in this file: [docs/standalone.md](https://github.com/valkey-io/libvalkey/blob/main/docs/standalone.md) including the options and macro.

There is no way to use MPTCP with the cluster API, such as `valkeyClusterConnectWithOptions`, right? There is a limitation in general that valkeyOptions can't be used with the cluster API. We can fix the cluster API later... @bjosv WDYT?

include/valkey/valkey.h
```diff
@@ -234,6 +238,14 @@ typedef struct {
         (opts)->endpoint.tcp.port = port_;       \
     } while (0)
 
+#define VALKEY_OPTIONS_SET_MPTCP(opts, ip_, port_) \
+    do {                                           \
+        (opts)->type = VALKEY_CONN_TCP;            \
+        (opts)->endpoint.tcp.ip = ip_;             \
+        (opts)->endpoint.tcp.port = port_;         \
+        (opts)->options = VALKEY_OPT_MPTCP;        \
```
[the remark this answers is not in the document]
pizhenwei: From my understood, `VALKEY_OPTIONS_SET_XXX` is used behind zeroed valkeyOptions, then apply other options after `VALKEY_OPTIONS_SET_XXX`, example codes in `valkey.c`:
```
valkeyContext *valkeyConnectNonBlock(const char *ip, int port) {
    valkeyOptions options = {0};
    VALKEY_OPTIONS_SET_TCP(&options, ip, port);
    options.options |= VALKEY_OPT_NONBLOCK;
    return valkeyConnectWithOptions(&options);
}

valkeyContext *valkeyConnectBindNonBlock(const char *ip, int port,
                                         const char *source_addr) {
    valkeyOptions options = {0};
    VALKEY_OPTIONS_SET_TCP(&options, ip, port);
    options.endpoint.tcp.source_addr = source_addr;
    options.options |= VALKEY_OPT_NONBLOCK;
    return valkeyConnectWithOptions(&options);
}

valkeyContext *valkeyConnectBindNonBlockWithReuse(const char *ip, int port,
                                                  const char *source_addr) {
    valkeyOptions options = {0};
    VALKEY_OPTIONS_SET_TCP(&options, ip, port);
    options.endpoint.tcp.source_addr = source_addr;
    options.options |= VALKEY_OPT_NONBLOCK | VALKEY_OPT_REUSEADDR;
    return valkeyConnectWithOptions(&options);
}
```

Please correct me if I misunderstand this rule.

pizhenwei:
> I think we can also mention it in the documentation.
> 
> * README under Features (somewhere near where we mention RDMA)

Done.

> * Documentation in this file: [docs/standalone.md](https://github.com/valkey-io/libvalkey/blob/main/docs/standalone.md) including the options and macro.
> 

Done.

@zuiderkwast thanks!

include/valkey/valkey.h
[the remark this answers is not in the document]
zuiderkwast: Is this rule documented?

The examples are just examples, not complete documentation.

I think it is better if it can work in any order. Otherwise some users can be confused. The macro names `VALKEY_OPTIONS_SET_XXX` indicates it sets some options, not removes other options.

Also it can be a good style like this:

```c
valkeyOptions opts = {.options = 
VALKEY_OPT_PREFER_IPV6};
VALKEY_OPTIONS_SET_MPTCP(&opts, "host.example", 12345);
```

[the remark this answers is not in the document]
pizhenwei: Right. Please take a look at the latest version.

zuiderkwast (approved):

pizhenwei:
Please squash there changes into a single commit, thanks.

michael-grunder:
Is this good to be merged, or are we holding off for the libvalkey devendoring.

bjosv:
> There is no way to use MPTCP with the cluster API, such as `valkeyClusterConnectWithOptions`, right? There is a limitation in general that valkeyOptions can't be used with the cluster API. We can fix the cluster API later... @bjosv WDYT?

I created #197 now to be able to enable MPTCP on the cluster API.

PR #190, opened 2025-04-09
Validate arguments for `%s`- and `%b`-format in `valkeyvFormatCommand`

bjosv:
- Make sure the expected `char*` argument given to `%s` or `%b` is a valid pointer.
  Avoids SIGSEGV in `sdscatlen`/`memcpy`.
  
- Output failure log when a test fails in CMake builds

### DIFF
--- a/src/valkey.c
+++ b/src/valkey.c
@@ -371,12 +371,16 @@ int valkeyvFormatCommand(char **target, const char *format, va_list ap) {
             switch (c[1]) {
             case 's':
                 arg = va_arg(ap, char *);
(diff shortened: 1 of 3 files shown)


michael-grunder (approved):
IIRC `printf("%s", NULL);` is UB so calling it a format error seems reasonable.

Other options:
* Printing `(null)` like many `printf` implementatiosn do. It would result in `(null)` values going into Valkey which is less than ideal.
* Aborting. Easier to debug but we don't really do that anywhere else.

PR #193, opened 2025-04-25
Minor corrections in tests and build of tests.

bjosv:
- CMake: append environment variables when running client_test.
    CMake will by default replace previously set variables which e.g. blocks us running tests towards redis-server with TLS.
- CMake: Build and run cluster tests that have required deps. and includes fulfilled (not if `LIBEVENT_LIBRARY` exists).
 - Remove redundant setup of test type (nit).
- Don't select a logical database when testing SSUBSCRIBE    
    SSUBSCRIBE requires a cluster node which doesn't support logical databases > 0 (Valkey will probably add support for it in later releases though).
-    Disable rdb- and aof-file writes in client_test to avoid interference between (failing) runs.


michael-grunder (approved):
LGTM but the CMake is greek to me

PR #196, opened 2025-04-28
Update and unify version tags on installed shared libraries

bjosv:
Previously a CMake-build and a pure Makefile-build gave different results when installing the built shared libraries.
This PR unifies the result, i.e. a CMake build and the provided Makefile gives the same results.

The PR also changes the version tags on installed libraries to include the patch version, and
the `SONAME` in the built libraries are changed to only include the major version as described in some [docs](https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html).

The naming of the version defines in `valkey.h` are improved, e.g. `LIBVALKEY_MAJOR` is changed to `LIBVALKEY_VERSION_MAJOR`, and the SONAME define is removed (MAJOR used instead..)

**Results after change:**

Linux (both Makefile and CMake):
```
libvalkey.so -> libvalkey.so.0
libvalkey.so.0 -> libvalkey.so.0.1.0
libvalkey.so.0.1.0

> readelf -a libvalkey.so | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libvalkey.so.0]
```
macOS (both Makefile and CMake):
```
libvalkey.dylib -> libvalkey.0.dylib
libvalkey.0.dylib -> libvalkey.0.1.0.dylib
libvalkey.0.1.0.dylib
```
**Results before change:**

Linux:
```
# Makefile
libvalkey.so -> libvalkey.so.0.1
libvalkey.so.0 -> libvalkey.so.0.1
libvalkey.so.0.1

# CMake
libvalkey.so -> libvalkey.so.0.1
libvalkey.so.0.1

> readelf -a libvalkey.so | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libvalkey.so.0.1]
```
macOS:
```
# Make
libvalkey.dylib -> libvalkey.0.1.dylib
libvalkey.0.dylib -> libvalkey.0.1.dylib
libvalkey.0.1.dylib

# CMake
libvalkey.dylib -> libvalkey.0.1.dylib
libvalkey.0.1.dylib
```


michael-grunder (approved):

zuiderkwast (approved):
Bumpin the major when incompatible changes doesn't apply to beta versions, major = 0. If we make incompatible changes without bumping it, it can break user code.

Let's reach 1.0 soon so we can avoid this problem?

bjosv:
> Let's reach 1.0 soon so we can avoid this problem?

Yes, that's a good reason to use 1.0.0 as our first tag.

PR #200, opened 2025-05-07
Remove macro UNUSED from public API

bjosv:
Update the examples, which uses the public API, but keep the macro in test only.
Additionally remove an unused define in `sds.c`.

### DIFF
--- a/examples/cluster-async-tls.c
+++ b/examples/cluster-async-tls.c
@@ -51,10 +51,7 @@ void disconnectCallback(const valkeyAsyncContext *ac, int status) {
     printf("Disconnected from %s:%d\n", ac->c.tcp.host, ac->c.tcp.port);
 }
 
-int main(int argc, char **argv) {
-    UNUSED(argc);
-    UNUSED(argv);
-
+int main(void) {
     valkeyTLSContext *tls;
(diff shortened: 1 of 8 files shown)


bjosv:
> Is main(void) backed by some standard?

Had to check, but its mentioned in the [C99](https://www.dii.uchile.cl/~daespino/files/Iso_C_1999_definition.pdf) standard and [N1256](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf)
```
5.1.2.2.1 Program startup
1 The function called at program startup is named main. The implementation declares no
prototype for this function. It shall be defined with a return type of int and with no
parameters:
int main(void) { /* ... */ }
or with two parameters (referred to here as argc and argv, though any names may be
used, as they are local to the function in which they are declared):
int main(int argc, char *argv[]) { /* ... */ }
or equivalent;9) or in some other implementation-defined manner.
```

michael-grunder:
There's a bunch of fun trivia around `main` specifically. It's the only place (that I know of) where it's well-defined to just not return a value even if it has a return type.

```c
// well defined in >= c99, UB in < c99
int main() { }
```

PR #201, opened 2025-05-08
Fix dependency issue with RDMA

bjosv:
- Make `vk_usec_now` inlined to avoid having `libvalkey_rdma.a` depend on `libvalkey.a` for this function (and getting link order issues).
- Build tests by linking to `libvalkey.a` before `libvalkey_rdma.a` to indicate dependency issues.


zuiderkwast (approved):
@pizhenwei FYI

PR #203, opened 2025-05-11
fix windows build

trim21:
I think it's expected to include `windows.h` before `wincrypt.h` if we have `WIN32_LEAN_AND_MEAN` defined


trim21:
obviously you should not reorder the includes of c/cxx files...


https://github.com/valkey-io/libvalkey/commit/60918c1f4ae38b0e6f2022b088e065b94564587c#diff-34d62efe7620eac9235c878aafc6716d1ef20a2246a35fac977f34bacce0481c

.github/workflows/ci.yml
```diff
@@ -239,10 +239,16 @@ jobs:
           sed -i '3i   set(ENV{CMAKE_POLICY_VERSION_MINIMUM} 3.5)'  ${env:VCPKG_INSTALLATION_ROOT}/scripts/ports.cmake
           sed -i '4i endif()'                                       ${env:VCPKG_INSTALLATION_ROOT}/scripts/ports.cmake
           vcpkg install --triplet x64-windows pkgconf libevent
+
+      - run: rm 'C:/Program Files/OpenSSL/' -rf
+        shell: bash
+
+      - run: choco install openssl
```
bjosv, line 246: You can remove this added section (including the `rm 'C:/...`) and just add openssl to line 236, 
i.e
`choco install -y memurai-developer openssl`

.github/workflows/ci.yml
```diff
@@ -239,10 +239,16 @@ jobs:
...
+
       - name: Build
         run: |
           mkdir build && cd build
-          cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=C:\vcpkg\scripts\buildsystems\vcpkg.cmake
+          cmake .. -G Ninja -D ENABLE_TLS=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=C:\vcpkg\scripts\buildsystems\vcpkg.cmake
```
bjosv, line 251:
```suggestion
          cmake .. -G Ninja -DENABLE_TLS=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=C:\vcpkg\scripts\buildsystems\vcpkg.cmake
```
nitpick, all other options are given as a single argument in this file.

bjosv (changes requested):
Nice findings!

I'm not sure why we need WIN32_LEAN_AND_MEAN, but I guess it works. Maybe you know?

trim21:
> Nice findings!
> 
> I'm not sure why we need WIN32_LEAN_AND_MEAN, but I guess it works. Maybe you know?

So `windows.h` doesn't include headers like MFC

.github/workflows/ci.yml
trim21: there is a openssl 1.1.1w at 'C:/Program Files/OpenSSL/`

bjosv: Ah, I see. Is the there any problem with default version already installed? Maybe you want to make sure a maintained version of openssl works?

trim21: We don't even enabled TLS on windows ci before this, so I think it's fine to just add latest openssl version?

bjosv (approved):

PR #205, opened 2025-05-13
Don't expose internal functions in shared libraries

### LINKED ISSUE #45: Don't expose internal functions in .so file
To avoid name collisions when linking projects with other versions of dict, sds, etc. we shouldn't expose these functions in the .so file.

We can compile with `-fvisibility=hidden` and explicitly add `__attribute__((visibility("default")))` to the exported functions. It's explained in https://stackoverflow.com/a/4466766/975856.

(Related to #44.)

bjosv:
- Add `-fvisiblity=hidden` flag in CMake and Makefile, i.e. all functions are hidden by default.
- Use a new macro `LIBVALKEY_API` to mark functions provided in the API.
- Unhide functions in `libvalkey.so` used by `libvalkey_tls.so`.
- Link unit-tests to a static library to avoid symbol hiding problems.
    When building shared libraries in CMake, i.e. `BUILD_SHARED_LIBS=ON` (default),
    a new static library `libvalkey_unittest.a` is created.
    This is used in unittests to access symbols that would be hidden in a shared library.
- Replace `sdsll2str`/`sdsull2str` with alternative from `valkey`.
    The newer functions used in valkey is stated to be faster, but the improvement in libvalkey is minimal, at least not slower. The benefit is to handle an optimization bug in gcc for `ppc64le` and `s390x` that triggers a `stringop-overflow` warning when building using `-O3` and `-fvisiblity=hidden`.

Findings:
- LIBVALKEY_API needs to be positioned before the function due to restrictions when building for Windows.

Fixes #45


bjosv:
An open question is if we should be able to disable the visibility flag using an CMake option/Make flag. Some projects seem to have this possibility.

michael-grunder (approved):
This is nice. I haven't done much with `-fvisibility=hidden` though so I'm not really familiar with edge cases.

CMakeLists.txt
```diff
@@ -189,11 +188,11 @@ IF(ENABLE_TLS)
         SET_PROPERTY(TARGET valkey_tls PROPERTY LINK_FLAGS "-Wl,-undefined -Wl,dynamic_lookup")
     ENDIF()
 
-    SET_TARGET_PROPERTIES(valkey_tls
-        PROPERTIES
-        WINDOWS_EXPORT_ALL_SYMBOLS TRUE
-        SOVERSION "${VERSION_MAJOR}"
-        VERSION "${VERSION}")
+    set_target_properties(valkey_tls PROPERTIES
+      C_VISIBILITY_PRESET hidden
```
zuiderkwast, line 192: What's the indentation rule? Is it 2 spaces after an open ( and 4 spaces otherwise?

include/valkey/alloc.h
```diff
@@ -81,11 +82,11 @@ static inline void vk_free(void *ptr) {
 
 #else
 
-void *vk_malloc(size_t size);
-void *vk_calloc(size_t nmemb, size_t size);
-void *vk_realloc(void *ptr, size_t size);
-char *vk_strdup(const char *str);
-void vk_free(void *ptr);
+LIBVALKEY_API void *vk_malloc(size_t size);
```
zuiderkwast, line 85: Please remind me, why are the malloc functions in the public API?

src/async_private.h
```diff
@@ -77,7 +78,7 @@ static inline void refreshTimeout(valkeyAsyncContext *ctx) {
     }
 }
 
-void valkeyAsyncDisconnectInternal(valkeyAsyncContext *ac);
-void valkeyProcessCallbacks(valkeyAsyncContext *ac);
+LIBVALKEY_API void valkeyAsyncDisconnectInternal(valkeyAsyncContext *ac);
+LIBVALKEY_API void valkeyProcessCallbacks(valkeyAsyncContext *ac);
```
zuiderkwast, line 82: Why are these visible?

zuiderkwast (commented):
LGTM. Some questions.

include/valkey/alloc.h
michael-grunder: Becaused you can [replace them with your own allocators](https://github.com/valkey-io/libvalkey/blob/f670346b9d28f5c957c7457112faca4731edd503/include/valkey/alloc.h#L50)

This is really useful if you want to use the library under the hood in some larger system that uses a custom allocator.

michael-grunder: Hiding them is fine by me. Looking through the header again I think I did it that way because then you can make the wrappers that call the custom allocators inline. 

I can't imagine the performance difference will be meaningful if we hide them.

zuiderkwast: Good point actually. We can keep them for inlining. :)

src/async_private.h
bjosv: These two, and `valkeySetError` are internal functions, but since `tls.c` and `libvalkey_tls.so` needs them I had to make them public. Maybe I should add a comment about it?

CMakeLists.txt
bjosv: The file is a mix of the older 4 spaces and the newer 2.
All recent changes uses 2 spaces and lowercase for commands.
There is no official guideline, but the cmake-mode.el gives 2.

[Zephyrs](https://docs.zephyrproject.org/latest/contribute/style/cmake.html) guideline is also mentioned sometimes which uses 2.

src/async_private.h
zuiderkwast: Ah, OK. Yeah why not.

PR #213, opened 2025-05-30
Spelling

jsoref:


.github/wordlist.txt
```diff
@@ -7,7 +7,7 @@ antirez
 api
 APIs
 ASYNC
-asyncronous
```
jsoref, line 10: I'm not sure what this file is doing, but as someone who maintains a spell-checking tool, the presence of this entry is a red flag.

I'm working from this report: https://github.com/check-spelling-sandbox/libvalkey/actions/runs/15354064163/attempts/1#summary-43209009081

Anyway, the last thing anyone wants is to put a misspelled word into a dictionary as it'll mean that a tool will not report that misspelled word.

tests/client_test.c
```diff
@@ -1034,7 +1034,7 @@ static void test_allocator_injection(void) {
     valkeyResetAllocators();
 }
 
-#define VALKEY_BAD_DOMAIN "idontexist-noreally.com"
+#define VALKEY_BAD_DOMAIN "nonexistent.example.com"
```
jsoref, line 1037: https://www.rfc-editor.org/rfc/rfc2606.html

michael-grunder: I'm fine with this change. The reason i did it the first way was because someone actually registered `idontexist.com` so the tests were failing.

michael-grunder (approved):
Looks good to me. No logic changes here as far as I can tell.

I do feel like maybe I need to go to detention though.

bjosv (approved):

PR #229, opened 2025-08-12
Fix duplicate Acks for RDMA events.

ruihong123:
The old logic may result in the RDMA event being acknowledged unexpectly in the following two scenarios. (1) ibv_get_cq_event get an EAGAIN error. (2) ibv_get_cq_event get one event but it may ack multiple times in the pollcq loop.

This commit has been reviewed from the valkey-server side, I just refactor those changes in libvalkey. Please refer to https://github.com/valkey-io/valkey/pull/2430 for a quicker review.


bjosv (approved):
You know this best, but it LGTM.

bjosv:
The CI for Windows is corrected on main, so merging this.
Is this fix also added in the valkey repo (deps/libvalkey)?

bjosv:
> Yes, it would be fine to bump this into valkey 9.

I'll bump it.

PR #233, opened 2025-08-21
Fix the long-blocking read for Valkey RDMA.

ruihong123:
During the valkey benchmark, when we set the data size as 16KB, the benchmark will be blocking on GET. The reason behind is that RDMA event is edge-triggered and hence every incoming data has to be read totally instead of read partially. We modify the valkeyBufferRead to enable a total read over the RDMA buffer when the conneciton type is RDMA. To realize this logic we need to keep read the buffer until the read result data size is smaller than the attempt read size.

In addition, we need to deal with a corner case for the logic above. If the message happened to be 16KB and the first read finishes all the available data, the second read will be triggered and block there until return an error. To solve this problem, we make read non-blocking except for the first read.

An alternative approach to solve this problem could be making the message sender chunk the message into 16KB blocks (equal to the stack buffer) and generate the PollIn signal every 16KB data arrival. This approach seems to also improve the performance of RDMA commands with very large data size.  According to our benchmark result. this pipelined RDMA transfer can improve "SET" command throughput by 1.4X. I am okay with both solutions.

Please take a look at this PR @pizhenwei.


pizhenwei:
> During the valkey benchmark, when we set the data size as 16KB, the benchmark will be blocking on GET. The reason behind is that RDMA event is edge-triggered and hence every incoming data has to be read totally instead of read partially. Otherwise, the benchmark will be blocked at the event loop . We modify the valkeyBufferRead to enable a total read over the RDMA buffer when the conneciton type is RDMA. To realize this logic we need to keep read the buffer until the read result data size is smaller than the attempt read size.
> 
> In addition, we need to deal with a corner case for the logic above. If the message happened to be 16KB and the first read finishes all the available data, the second read will be triggered and block there until return an error. To solve this problem, we make read non-blocking except for the first read.
> 
> An alternative approach to solve this problem could be making the message sender chunk the message into 16KB blocks (equal to the stack buffer) and generate the PollIn signal every 16KB data arrival. This approach seems to also improve the performance of RDMA commands with very large data size. According to our benchmark result. this pipelined RDMA transfer can improve "SET" command throughput by 1.4X. I am okay with both solutions.
> 
> Please take a look at this PR @pizhenwei.

Hi
Thanks for your work!
I'm out of cycles this week, I'll dive into this issue next week.

src/valkey.c
```diff
@@ -1009,13 +1009,41 @@ int valkeyBufferRead(valkeyContext *c) {
...
+        }
+        if (nread > 0 && valkeyReaderFeed(c->reader, buf, nread) != VALKEY_OK) {
+            valkeySetError(c, c->reader->err, c->reader->errstr);
+            return VALKEY_ERR;
+        }
+    } else {
+        /* For Valkey over RDMA, as the ready-to-read signal only exists per arrived RDMA write.
+         * This function has to copy all the data from communication buffer to client buffer.*/
+        struct timeval *recorded_timeout = c->command_timeout;
+        struct timeval zero_timeout;
```
michael-grunder, line 1025:
```suggestion
        static struct timeval zero_timeout = {0};
```

Since this never changes you can use static storage

src/valkey.c
```diff
@@ -1009,13 +1009,41 @@ int valkeyBufferRead(valkeyContext *c) {
...
+        while (1) {
+            nread = c->funcs->read(c, buf, sizeof(buf));
+            if (nread < 0) {
+                return VALKEY_ERR;
+            }
+            if (nread > 0 && valkeyReaderFeed(c->reader, buf, nread) != VALKEY_OK) {
+                valkeySetError(c, c->reader->err, c->reader->errstr);
+                return VALKEY_ERR;
```
michael-grunder, line 1033: Same as above

ruihong123: Yeah, I agree.

ruihong123:
Hi @pizhenwei, I am wondering when we should update the internal offset (`ctx->recv_offset`) for the RDMA buffer for `c->funcs->read_zc`. I feel like it is 100% safe to update `ctx->recv_offset` within `c->funcs->read_zc`, unless the invoker is guaranteed to copy/use the data before calling next `connRdmaHandleCq` .  Do you think it is okay to update `ctx->recv_offset` within `c->funcs->read_zc`?

Thanks

pizhenwei:
> Hi @pizhenwei, I am wondering when we should update the internal offset (`ctx->recv_offset`) for the RDMA buffer in `c->funcs->read_zc`. I feel like it is not 100% safe to update `ctx->recv_offset` within `c->funcs->read_zc`, unless the invoker is guaranteed to copy/consume the data before calling next `connRdmaHandleCq` . Do you think it is okay to move forward `ctx->recv_offset` within `c->funcs->read_zc`?
> 
> Thanks

There are two jobs in current `valkeyRdmaRead`:
* move `ctx->recv_offset`. This is safe in `c->funcs->read_zc`
* may `connRdmaRegisterRx`. This will tell the remote side to a new buffer, generally it's the original one, the remote side may overwrite it during the local side is reading it.

So I suggest:
* just return the reading buffer pointer and length in `c->funcs->read_zc`
* update `ctx->recv_offset` and `connRdmaRegisterRx`(if necessary) in `c->funcs->read_zc_done`

Hi @michael-grunder do you have any suggestions?

michael-grunder:
@pizhenwei  You know much more about the RDMA feature than I do. I'm fine with your suggestion assuming it's safe.

ruihong123:
Hi everyone, the update is submitted. Please check it. Thanks.

ruihong123:
> Hi @ruihong123 Generally LGTM, thanks!
> 
> By the way, would you please add a few test result between the two version, I'm looking forward to seeing the improvement.

For some reasons, I lost the accesses for the testbed that I had the results of the first version. If you deem it necessary for merge, I can create a cloudlab cluster and compare the performance results there. However, it may take some time.

pizhenwei:
> > Hi @ruihong123 Generally LGTM, thanks!
> > By the way, would you please add a few test result between the two version, I'm looking forward to seeing the improvement.
> 
> For some reasons, I lost the accesses for the testbed that I had the results of the first version. If you deem it necessary for merge, I can create a cloudlab cluster and compare the performance results there. However, it may take some time.

The test report will not block merge code, do you test these code? I think this PR should be tested in your developing environment at least.

ruihong123:
Okay, I got it. I will test this new commit in cloudlab in the next few days. We can merge this PR after I get the experiment result.

ruihong123:
Hi here is the benchmark result over the cloudlab.
+-------------+-----------------------------+-----------------------------+-----------------------------+-----------------------------+
| Data Size   | SET Before Optimization     | SET After Optimization      | GET Before Optimization     | GET After Optimization      |
+-------------+-----------------------------+-----------------------------+-----------------------------+-----------------------------+
| 3 B         | 199,362.05                  | 191,644.31                  | 212269.16                  | 218197.67                  |
| 1024 B      | 152,718.39                  | 150,943.39                  | 189825.36                  | 198,846.70                  |
| 16,384 B    | 36,064.63                   | 36,709.37                   | 84,076.01                   | 118,652.11                  |
| 1,048,576 B | 418.88                      | 413.33                      | NA                           | 74,626.87                   |
+-------------+-----------------------------+-----------------------------+-----------------------------+-----------------------------+

The zero-copy implementation can function properly with varied data sizes, which fixes the blocking issue for this PR. With 16KB, the zero-copy optimization can bring in ~1.41X throughput boost, and the performance gain could be higher as the data size grows. However, the data points after 16KB are not available for GET before the optimization. 

By the way, there is a huge throughput drop for the SET operation with 1MB data size. I recall seeing a similar issue with TCP connections, so I think this issue primarily results from memory allocation.

pizhenwei:
> Hi here is the benchmark result over the cloudlab C6220 nodes. SET and GET command throughputs (ops/sec) are shown with/without the optimization. +-------------+-----------------------------+-----------------------------+-----------------------------+-----------------------------+ | Data Size | SET Before Optimization | SET After Optimization | GET Before Optimization | GET After Optimization | +-------------+-----------------------------+-----------------------------+-----------------------------+-----------------------------+ | 3 B | 199,362.05 | 191,644.31 | 212269.16 | 218197.67 | | 1024 B | 152,718.39 | 150,943.39 | 189825.36 | 198,846.70 | | 16,384 B | 36,064.63 | 36,709.37 | 84,076.01 | 118,652.11 | | 1,048,576 B | 418.88 | 413.33 | NA | 74,626.87 | +-------------+-----------------------------+-----------------------------+-----------------------------+-----------------------------+
> 
> The zero-copy implementation can function properly with varied data sizes, which fixes the blocking issue for this PR. With 16KB, the zero-copy optimization can bring in ~1.41X throughput boost, and the performance gain could be higher as the data size grows. However, the data points after 16KB are not available for GET before the optimization.
> 

16K ~ 32K KV size is quite common in the real workloads, great jobs! thanks!

> By the way, there is a huge throughput drop for the SET operation with 1MB data size. I recall seeing a similar issue with TCP connections, so I think this issue primarily results from memory allocation.

This should be another issue, so I think we can merge this PR, then fix it in another PR.

src/rdma.c
```diff
@@ -540,6 +540,46 @@ static ssize_t valkeyRdmaRead(valkeyContext *c, char *buf, size_t bufcap) {
...
+        return VALKEY_ERR;
+    }
+}
+
+static void valkeyRdmaReadZCDone(valkeyContext *c) {
+    RdmaContext *ctx = c->privctx;
+    struct rdma_cm_id *cm_id = ctx->cm_id;
+
+    if (ctx->recv_offset == ctx->recv_length) {
+        connRdmaRegisterRx(c, cm_id);
```
bjosv, line 579: I don't know the RDMA feature, but `connRdmaRegisterRx` seems to be able to fail when it calls `rdmaSendCommand` ("no empty command buffer" or if `ibv_post_send` fails).

Could there be any problems with the client got stuck/hanging in this case?
Is there a need for a return code in `read_zc_done` to indicate an unrecoverable error?

src/rdma.c
```diff
@@ -991,6 +1031,8 @@ static valkeyContextFuncs valkeyContextRdmaFuncs = {
     .async_read = valkeyRdmaAsyncRead,
     .async_write = valkeyRdmaAsyncWrite,
     .read = valkeyRdmaRead,
```
bjosv, line 1033: Should we replace the handling within `valkeyRdmaRead` with an assert to avoid having unused code?

pizhenwei: Good idea, no need to keep the dead code.

michael-grunder (approved):

include/valkey/valkey.h
```diff
@@ -270,6 +270,12 @@ typedef struct valkeyContextFuncs {
      * these functions shall return a value < 0.  In the event of a
      * recoverable error, they should return 0. */
     ssize_t (*read)(struct valkeyContext *, char *, size_t);
+    /* ZC means zero copy, it provides underlay transport layer buffer directly,
+     * so it has better performance than generic read. After consuming the read
+     * buffer, it's necessary to notify the underlay transport to advance the
+     * read buffer by read_zc_done. */
+    ssize_t (*read_zc)(struct valkeyContext *, char **);
+    ssize_t (*read_zc_done)(struct valkeyContext *);
```
bjosv: Using an int here would better describe that only a status code is returned.
WDYT?
```suggestion
    int (*read_zc_done)(struct valkeyContext *);
```

Same below in `static ssize_t valkeyRdmaReadZCDone`

pizhenwei: I notice this function family also use ssize_t, so I think ssize_t for read_zc_done will be better.
```
    ssize_t (*read)(struct valkeyContext *, char *, size_t);
    ssize_t (*write)(struct valkeyContext *);
```

PR #244, opened 2025-09-29
Add option to select a logical database

bjosv:
Set the option `select_db` in `valkeyClusterOptions` to select a logical database after a successful connect.
The option is supported in both the Synchronous and Asynchronous API, and the `SELECT` command is only sent when configured to a non-zero value.

Additionally adds new flags  to the clusterclients `[--select-db NUM]` and update tests.


bjosv:
> We don't have anything like this for standalone though. There, users have to do it themselves in the connect callback? Can we do something more convenient there?

Interesting idea.
@michael-grunder Do you remember if there has been any talks about adding post-`connect` procedures with `AUTH` and `SELECT` as options? It would make it easier for users, but maybe there are reasons?

Issue #247, opened 2025-10-07
linker errors with cmake and TLS/RDMA support


andreasstieger:
Building 0.2.1 with cmake and one of TLS or RDMA support results in linker errors. This is on openSUSE Tumbleweed and openSUSE Leap 16.0. 

* known not working: cmake 4.1.2 / gcc 15
* known not working: cmake 3.31.7 / gcc 15
* known working: cmake 3.28.3 / gcc 7
* known working: make / gcc 15

Full logs (currently) in https://build.opensuse.org/package/show/home:AndreasStieger/libvalkey. Partial logs below.

Turning on only RDMA and disabling TLS will result in equivalent errors linking the RDMA library.

````
[    6s] [ 47%] Linking C shared library libvalkey_tls.so
[    6s] /usr/bin/cmake -E cmake_link_script CMakeFiles/valkey_tls.dir/link.txt --verbose=1
[    6s] /usr/lib64/gcc/x86_64-suse-linux/15/../../../../x86_64-suse-linux/bin/ld: /tmp/ccrs4npl.ltrans0.ltrans.o: in function `valkeyTLSConnect':
[    6s] /home/abuild/rpmbuild/BUILD/libvalkey-0.2.1-build/libvalkey-0.2.1/include/valkey/alloc.h:68:(.text+0x30): undefined reference to `valkeyAllocFns'
[    6s] /usr/lib64/gcc/x86_64-suse-linux/15/../../../../x86_64-suse-linux/bin/ld: /tmp/ccrs4npl.ltrans0.ltrans.o: in function `valkeyTLSConnect':
[    6s] /home/abuild/rpmbuild/BUILD/libvalkey-0.2.1-build/libvalkey-0.2.1/src/tls.c:403:(.text+0x130): undefined reference to `valkeySetError'
[    6s] /usr/lib64/gcc/x86_64-suse-linux/15/../../../../x86_64-suse-linux/bin/ld: /tmp/ccrs4npl.ltrans0.ltrans.o: in function `valkeyTLSConnect':
[    6s] /home/abuild/rpmbuild/BUILD/libvalkey-0.2.1-build/libvalkey-0.2.1/include/valkey/alloc.h:80:(.text+0x137): undefined reference to `valkeyAllocFns'
[...]

PR #248, opened 2025-10-08
Fix linker errors when building with cmake, TLS and -Wl,--no-undefined

andreasstieger:
Part of #247


michael-grunder:
This seems to cause an ASAN issue (`valkeyAllocFns` being defined in both `client_test` and `libvalkey.so`).

I'll have to verify locally to confirm.

bjosv:
@michael-grunder is this PR good enough?
My plan is to fix the Makefile in another PR together with some cleanups.

michael-grunder:
My cmake is limited, but looks good to me

PR #294, opened 2026-03-17
Optimize read buffer compaction and reduce copying

zuiderkwast:
Do compaction of the read buffer (memmove) before feeding more data to the reader, instead of after each parsed command.

Previously, sdsrange() was called in valkeyReaderGetReply() to compact the read buffer after consuming >= 1024 bytes. With pipelining, GetReply is called many times per Feed call (once per reply), causing repeated memmove() operations on the remaining buffer data.

Move the compaction to valkeyReaderFeed(), right before appending new data. This ensures the memmove happens at most once per network read instead of once per parsed reply.

Additionally, make space in the allocated read buffer before the read() call and read directly into the allocated reply buffer instead of copying it via a static buffer. A public API for this is added (valkeyReaderGetReadBuf + valkeyReaderCommitRead).

Profiling
---------

Profiled using valkey-benchmark on macOS (Apple M4 Pro) with:

    valkey-benchmark -P 32 -d 1024 -r 1000000 -n 3000000 -c 50 -t set,get

Server ran with --io-threads 4 to avoid being server-bottlenecked.

Flamegraph sampling (macOS `sample` tool, 10s at 1ms intervals) of the valkey-benchmark process shows the following top userspace hotspots:

| Read path hotspot | Original | After compaction fix | After direct read |
|---|---|---|---|
| sdsrange → memmove (GetReply) | 250 | 0 ← eliminated | 0 |
| valkeyReaderFeed → sdscatlen → memmove | 38 | 43 | 0 ← eliminated |
| __recvfrom (kernel) | 246 | 266 | 295 |
| createStringObject → memmove | 25 | 27 | 28 |

After the first change, the sdsrange memmove (250 samples) is eliminated and the compaction cost is absorbed into valkeyReaderFeed with negligible overhead increase (38 -> 43 samples). After the second change, the sdscatlen -> memmove is also eliminated.

### DIFF
--- a/src/read.c
+++ b/src/read.c
@@ -790,6 +790,57 @@ oom:
     return VALKEY_ERR;
 }
 
+/* Prepare the reader's internal buffer for a direct read. This compacts
+ * consumed data, ensures at least 'minbytes' of writable space, and returns
+ * a pointer and available capacity. The caller can then read() directly into
+ * *buf and call valkeyReaderCommitRead() with the number of bytes read.
+ * Returns VALKEY_OK on success, VALKEY_ERR on allocation failure. */
+int valkeyReaderGetReadBuf(valkeyReader *r, char **buf, size_t *cap, size_t minbytes) {
+    if (r->err)
+        return VALKEY_ERR;
+
+    /* Destroy internal buffer when it is empty and is quite large. */
+    if (r->len == 0 && r->maxbuf != 0 && sdsavail(r->buf) > r->maxbuf) {
+        sdsfree(r->buf);
+        r->buf = sdsempty();
+        if (r->buf == NULL)
+            goto oom;
+        r->pos = 0;
+    }
+
+    /* Compact consumed data. */
+    if (r->pos > 0) {
+        if (sdslen(r->buf) > SSIZE_MAX)
+            goto oom;
+        sdsrange(r->buf, r->pos, -1);
+        r->pos = 0;
+        r->len = sdslen(r->buf);
+    }
+
+    /* Ensure enough writable space. */
+    if (sdsavail(r->buf) < minbytes) {
+        sds newbuf = sdsMakeRoomFor(r->buf, minbytes);
+        if (newbuf == NULL)
+            goto oom;
+        r->buf = newbuf;
+    }
+
+    *buf = r->buf + sdslen(r->buf);
+    *cap = sdsavail(r->buf);
+    return VALKEY_OK;
+
+oom:
+    valkeyReaderSetErrorOOM(r);
+    return VALKEY_ERR;
+}
+
+/* Commit bytes that were written directly into the buffer obtained from
+ * valkeyReaderGetReadBuf(). */
+void valkeyReaderCommitRead(valkeyReader *r, size_t nread) {
+    sdsIncrLen(r->buf, (int)nread);
+    r->len = sdslen(r->buf);
+}
+
 int valkeyReaderGetReply(valkeyReader *r, void **reply) {
     /* Default target pointer to NULL. */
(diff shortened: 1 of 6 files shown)


zuiderkwast:
I added another optimization: direct read into the allocated reply buffer instead of via a static buffer.

| Read path hotspot | Original | After compaction fix | After direct read |
|---|---|---|---|
| sdsrange → memmove (GetReply) | 250 | 0 ← eliminated | 0 |
| valkeyReaderFeed → sdscatlen → memmove | 38 | 43 | 0 ← eliminated |
| __recvfrom (kernel) | 246 | 266 | 295 |
| createStringObject → memmove | 25 | 27 | 28 |

The sdscatlen copy in the read path (previously 38-43 samples) is now zero — data goes directly from the kernel into the reader's sds buffer. The kernel 
syscall samples went up proportionally since it's now a larger share of the remaining work.

The two optimizations combined eliminated the top two userspace hotspots in the read path. What remains is dominated by syscalls and the inherent cost of 
creating reply objects.

src/read.c
```diff
@@ -779,6 +790,57 @@ int valkeyReaderFeed(valkeyReader *r, const char *buf, size_t len) {
...
+
+oom:
+    valkeyReaderSetErrorOOM(r);
+    return VALKEY_ERR;
+}
+
+/* Commit bytes that were written directly into the buffer obtained from
+ * valkeyReaderGetReadBuf(). */
+void valkeyReaderCommitRead(valkeyReader *r, size_t nread) {
+    sdsIncrLen(r->buf, (int)nread);
```
michael-grunder: Do we want to be paranoid about the `size_t` -> `int` cast?

src/read.c
```diff
@@ -765,6 +765,17 @@ int valkeyReaderFeed(valkeyReader *r, const char *buf, size_t len) {
             r->pos = 0;
         }
 
+        /* Discard consumed data before appending, to avoid unbounded growth.
+         * This replaces the per-reply sdsrange() in valkeyReaderGetReply(),
+         * so the memmove happens at most once per feed call. */
+        if (r->pos > 0) {
+            if (sdslen(r->buf) > SSIZE_MAX)
+                goto oom;
```
bjosv: `goto oom;` will set an "Out of memory" error string, maybe an `return VALKEY_ERR;` is enough when we check sdslen()? (same in valkeyReaderGetReadBuf below in that case)

src/read.c
zuiderkwast: Yeah, we shouldn't ignore it.

I checked in valkey's sds. There, this parameter of sdsIncrLen is `ssize_t`. It has been like that since 2017. :)

We should probably do the same change in libvalkey's sds.

src/read.c
zuiderkwast: When will sdslen return SSIZE_MAX or larger? 32-bit platform and sdslen > 2G?

michael-grunder: Probably a very unlikely edge case but possible? `sdslen` returns a `size_t`.

Another option on this line is to check the `sdsrange` return value as it also checks if `sdslen(s) > SSIZE_MAX`.

zuiderkwast: An error should aways be set when we return VALKEY_ERR.

My assistant says:

> You're right, that's misleading. The check guards against sdsrange receiving a value that overflows ssize_t, which is a protocol/overflow error, not OOM.

```suggestion
                valkeyReaderSetError(r, VALKEY_ERR_PROTOCOL,
                                     "Reader buffer is too large");
                return VALKEY_ERR;
```

michael-grunder (approved):

bjosv (approved):

PR #297, opened 2026-03-23
Rename `freeReplyObject` to `valkeyFreeReplyObject`

bjosv:
The old name was inherited from hiredis and conflicts when both libraries are linked into the same binary.
Rename to follow the valkey prefix convention used by the rest of the API.
    
This is a breaking API change.

Fixes #288

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,10 @@ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_TLS=1 -DENABLE_RDMA=1 ..
 sudo make install
 ```
 
+### Other build options
+
+`VALKEY_NO_DEPRECATED` — Disable deprecated names (e.g. `freeReplyObject`). Define this to ensure only the new `valkey`-prefixed names are used.
+
 ## Contributing
 
 Please see [`CONTRIBUTING.md`](https://github.com/valkey-io/libvalkey/blob/main/CONTRIBUTING.md).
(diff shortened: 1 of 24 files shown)


zuiderkwast (commented):
This function is used by all users so it has a large blast radius, right? Annoying for those who already migrated.

Can we add a define for compatibility?

```c
#define freeReplyObject valkeyFreeReplyObject
```

You'd only get it if you include valkey.h so it doesn't conflict for other parts of the system. Maybe also guard it with some ifdef. WDYT?

michael-grunder:
IMO it's probably worth making this the default behavior when we release v1.0.0.

README.md
```diff
@@ -64,6 +64,10 @@ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_TLS=1 -DENABLE_RDMA=1 ..
 sudo make install
 ```
 
+### Other build options
+
+`VALKEY_NO_DEPRECATED` — Disable deprecated names (e.g. `freeReplyObject`). Define this to ensure only the new `valkey`-prefixed names are used.
```
zuiderkwast, line 69: It's not a make variable like the other build options mentioned above. Can we worth mentioning. How do you specify defines on the command line? Like this?

    make `CFLAGS=-DVALKEY_NO_DEPRECATED=1`

zuiderkwast (commented):
Shall we mention this rename in the migration guide?

PR #300, opened 2026-04-13
Try all addresses from DNS before failing to connect

bjosv:
When a hostname resolves to multiple addresses, libvalkey now iterates through all of them before giving up.
Previously only `EHOSTUNREACH` would try the next address; all other connect errors failed immediately.

The connect_timeout applies per address, not to the overall attempt.
With N addresses that all time out, the total wait is up to N * timeout.

The first commit refactors the connect error handling to prepare for the change:
- Removes the confusing `wait_for_ready` goto label
- Fixes `TCP_NODELAY` not being set on immediate connect success
- Reorders `errno` checks to put `EINPROGRESS` first
- Makes each error branch self-contained

The second commit adds the fallback behavior and a unit test that overrides `getaddrinfo` and `connect` to verify that the library falls back to a working address after earlier ones fail with `ECONNREFUSED`.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -561,6 +561,7 @@ int valkeyContextConnectTcp(valkeyContext *c, const valkeyOptions *options) {
 
         c->flags |= VALKEY_CONNECTED;
         rv = VALKEY_OK;
+        valkeyClearError(c);
         goto end;
     }
     if (p == NULL) {
--- a/src/net.c
+++ b/src/net.c
@@ -534,23 +534,21 @@ int valkeyContextConnectTcp(valkeyContext *c, const valkeyOptions *options) {
 
         if (connect(s, p->ai_addr, p->ai_addrlen) == -1) {
             if (errno == EINPROGRESS) {
-                if (blocking && valkeyContextWaitReady(c, timeout_msec) != VALKEY_OK)
-                    goto error;
+                if (blocking && valkeyContextWaitReady(c, timeout_msec) != VALKEY_OK) {
+                    valkeyNetClose(c);
+                    continue;
+                }
                 /* Non-blocking: EINPROGRESS is expected, continue to success. */
-            } else if (errno == EHOSTUNREACH) {
-                valkeyNetClose(c);
-                continue;
             } else if (errno == EADDRNOTAVAIL && reuseaddr) {
                 valkeyNetClose(c);
                 if (++reuses >= VALKEY_CONNECT_RETRIES) {
-                    goto error;
+                    continue;
                 } else {
                     goto addrretry;
                 }
             } else {
-                valkeySetErrorFromErrno(c, VALKEY_ERR_IO, NULL);
                 valkeyNetClose(c);
-                goto error;
+                continue;
             }
         }
         /* TCP_NODELAY is set here for blocking connections only. For non-blocking,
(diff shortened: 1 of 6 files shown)


bjosv:
When digging in hiredis history the handling of `EHOSTUNREACH` was added for dual-stack systems where getaddrinfo returns an IPv6 address before IPv4. If the host has no IPv6 connectivity, `connect()` to the IPv6 address fails with `EHOSTUNREACH`, c4ed06d fixed that only case.

With this change we support usecases when running in K8s and connecting to a service.

src/net.c
```diff
@@ -532,32 +533,29 @@ int valkeyContextConnectTcp(valkeyContext *c, const valkeyOptions *options) {
...
 
         if (connect(s, p->ai_addr, p->ai_addrlen) == -1) {
-            if (errno == EHOSTUNREACH) {
-                valkeyNetClose(c);
-                continue;
-            } else if (errno == EINPROGRESS) {
-                if (blocking) {
-                    goto wait_for_ready;
+            if (errno == EINPROGRESS) {
+                if (blocking && valkeyContextWaitReady(c, timeout_msec) != VALKEY_OK) {
```
michael-grunder, line 537: Since `valkeyContextWaitReady` will call `valkeySetError`, could a subsequent address successfully connect but leave the context in an error state?

e.g. the first `valkeyContextWaitReady` fails, we continue, and then a subsequent address succeeds.

src/net.c
```diff
@@ -566,9 +564,7 @@ int valkeyContextConnectTcp(valkeyContext *c, const valkeyOptions *options) {
         goto end;
```
michael-grunder, line 565: We could add a helper to clear the error somewhere on this success path?

```suggestion
        valkeyClearError(c);
        goto end;
```

michael-grunder:
This is much nicer. That jump to `wait_for_ready` _inside_ an `else` always confused me :smile:

src/net.c
bjosv: ah! Good catch, I will take a look and fix

src/net.c
bjosv: Yes, that sounds like a plan

michael-grunder (approved):

src/valkey.c
```diff
@@ -709,6 +709,11 @@ void valkeySetError(valkeyContext *c, int type, const char *str) {
     }
 }
 
+void valkeyClearError(valkeyContext *c) {
+    c->err = 0;
+    memset(c->errstr, '\0', strlen(c->errstr));
+}
```
zuiderkwast, line 715: We do this in various places. Should we use the function always?

src/valkey.c
```diff
@@ -709,6 +709,11 @@ void valkeySetError(valkeyContext *c, int type, const char *str) {
     }
 }
```
zuiderkwast, line 710: I see that valkeySetErrorFromErrno is very similar to this special case. We could sort out the different cases, but it might not be the scope of this PR.

zuiderkwast (approved):

src/valkey.c
bjosv: Not that many actually, but it would be good to use this function always, like in `valkeyReconnect`.

src/valkey.c
bjosv: Ah, yes, good idea. Another PR where we also use valkeyClearError as mentioned in the other comment.

PR #301, opened 2026-04-15
RDMA: Fix lost EPOLLIN/POLLIN events

pizhenwei:
There is only EPOLLIN/POLLIN events in RDMA, when an event arrives, it indicates that 1 or more CQ completions:
- IBV_WC_RECV: RegisterXferMemory command refreshes TX buffer, it equivalents to EPOLLOUT/POLLOUT for upper layer
- IBV_WC_RECV_RDMA_WITH_IMM: new payload arrives, it equivalents to EPOLLIN/POLLIN for upper layer ...

During *writing* data to the remote side, the client side may handle the completions like:
NO enought TX buffer
  -> wait IBV_WC_RECV_RDMA_WITH_IMM
    -> wait IBV_WC_RECV event
      -> write data

In such scenario, the client side consumes the
*IBV_WC_RECV_RDMA_WITH_IMM* in the *writing* method, and the outside event loop loses it. To avoid this issue, use a *eventfd* to kick. There are 3 FD(CM channel, completion channel, eventfd), so use a nested epollfd to hide the detail to the uplayer.

### DIFF
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -48,6 +48,8 @@
 #include <poll.h>
 #include <rdma/rdma_cma.h>
 #include <stdint.h>
+#include <sys/epoll.h>
+#include <sys/eventfd.h>
 #include <sys/socket.h>
 #include <sys/types.h>
 #include <unistd.h>
@@ -258,6 +260,7 @@ typedef struct RdmaContext {
     struct ibv_comp_channel *comp_channel;
     struct ibv_cq *cq;
     struct ibv_pd *pd;
+    int evfd;
 
     /* TX */
     char *tx_addr;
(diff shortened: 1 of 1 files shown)


src/rdma.c
```diff
@@ -969,13 +1046,20 @@ static int valkeyRdmaWaitConn(valkeyContext *c, long timeout) {
...
+        nevent = epoll_wait(c->fd, events, sizeof(events) / sizeof(events[0]), end - now);
+        if (!nevent) {
+            break;
         }
 
+        if (nevent < 0) {
+            if (errno == EINTR) {
+                continue;
+            }
+            break;
```
bjosv: This is an unexpected error from epoll_wait I guess, we don't need to valkeySetError here?
This `break;` could optionally also be replaced with a return VALKEY_ERR for clarity.

src/rdma.c
```diff
@@ -609,48 +658,61 @@ static int connRdmaHandleCq(valkeyContext *c) {
...
-
-    /* pfd[0] for CM event */
-    pfd[VALKEY_RDMA_POLLFD_CM].fd = ctx->cm_channel->fd;
-    pfd[VALKEY_RDMA_POLLFD_CM].events = POLLIN;
-    pfd[VALKEY_RDMA_POLLFD_CM].revents = 0;
+    while (1) {
+        nevent = epoll_wait(c->fd, events, sizeof(events) / sizeof(events[0]), timed);
+        if (nevent > 0) {
+            break;
+        } else if (!nevent) {
```
bjosv: optionally; `else` here and on line 684 could be removed for readability

bjosv (commented):
LGTM, not that familiar. Just some nits/questions that might be interesting to address before merging

pizhenwei:
Hi @bjosv 
Force pushed a new version, thanks for your suggestions.

bjosv (approved):

PR #302, opened 2026-04-16
Remove support for external dict while still supporting external sds

zuiderkwast:
These two appeared independent, but because they both set include paths, they couldn't be used independently. Instead of fixing that, we just remove the support for using external dict.

The reason is that valkey (where libvalkey is vendored and uses external components from valkey's counterparts) has changed the dict implementation so that the API is no longer compatible, but there are no symbol collisions either, so it's fine for libvalkey to use its own dict.

Extracted from
https://github.com/valkey-io/valkey/commit/fb655dbf5cfaf254e0b8bb6617f1f734dbebc07d


Makefile
bjosv:
```suggestion
# Allow the libvalkey provided sds type to be replaced by
```
suggestion got weird.. but remove dict in comment

bjosv (approved):
Nice!

PR #303, opened 2026-04-16
Refactor error handling to use valkeySetErrorFromErrno and valkeyClearError

bjosv:
Clean up error handling across the codebase:

- Promote `valkeySetErrorFromErrno()` from a static function in `net.c` to an exported function in `valkey.c`, making it available to `tls.c`.
- Remove the NULL special case from `valkeySetError()` since all callers now use `valkeySetErrorFromErrno()` for `errno`-based errors.
- Replace thread-unsafe `strerror(errno)` calls with `valkeySetErrorFromErrno()` which uses `strerror_r()` internally.
- Use `valkeyClearError()` in `valkeyReconnect()`.

### DIFF
--- a/src/valkey_private.h
+++ b/src/valkey_private.h
@@ -42,6 +42,7 @@
 #include <string.h>
 
 LIBVALKEY_API void valkeySetError(valkeyContext *c, int type, const char *str);
+LIBVALKEY_API void valkeySetErrorFromErrno(valkeyContext *c, int type, const char *prefix);
 void valkeyClearError(valkeyContext *c);
 
 /* Helper function. Convert struct timeval to millisecond. */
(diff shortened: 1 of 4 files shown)


src/valkey_private.h
```diff
@@ -42,6 +42,7 @@
 #include <string.h>
 
 LIBVALKEY_API void valkeySetError(valkeyContext *c, int type, const char *str);
+LIBVALKEY_API void valkeySetErrorFromErrno(valkeyContext *c, int type, const char *prefix);
```
zuiderkwast, line 45: Q: Why do we use this qualifier for internal API?

zuiderkwast (approved):
Nice cleanup.

src/valkey_private.h
bjosv: Its needed to be exported in the libvalkey.so so that libvalkey_tls.so can use it.
Maybe we could have them as static inline:d instead, but then duplicated.

bjosv: valkeySetError is also used in libvalkey_rdma.so.

PR #305, opened 2026-04-20
Fix potential heap-buffer-overflow in cluster error reply parsing

bjosv:
Add length checks in getReplyErrorType() before calling memcmp to prevent reading past the end of reply->str when a server sends a short error reply.

While a conforming server always sends complete error strings like "MOVED 1234 host:port", a malicious or buggy server could send a truncated error (e.g., -MO\r\n), causing memcmp to read beyond the allocated buffer. This is undefined behavior and is flagged by AddressSanitizer.

Also update the sync test client (clusterclient.c) to print server error replies with an "Error:" prefix, distinguishing them from successful replies, and add a test case that exercises the fix.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -212,13 +212,13 @@ static replyErrorType getReplyErrorType(valkeyReply *reply) {
 
     if (reply->type != VALKEY_REPLY_ERROR)
         return CLUSTER_NO_ERROR;
-    if (memcmp(reply->str, "MOVED", 5) == 0)
+    if (reply->len >= 5 && memcmp(reply->str, "MOVED", 5) == 0)
         return CLUSTER_ERR_MOVED;
-    if (memcmp(reply->str, "ASK", 3) == 0)
+    if (reply->len >= 3 && memcmp(reply->str, "ASK", 3) == 0)
         return CLUSTER_ERR_ASK;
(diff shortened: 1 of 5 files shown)


src/cluster.c
```diff
@@ -212,13 +212,13 @@ static replyErrorType getReplyErrorType(valkeyReply *reply) {
 
     if (reply->type != VALKEY_REPLY_ERROR)
         return CLUSTER_NO_ERROR;
-    if (memcmp(reply->str, "MOVED", 5) == 0)
+    if (reply->len >= 5 && memcmp(reply->str, "MOVED", 5) == 0)
```
zuiderkwast, line 215: Just an idea. Would it make sense to introduce a little helper function for this?

```C
static int replyErrorStartsWith(valkeyReply *reply, const char *prefix) {
    size_t prefix_len = strlen(prefix);
    return (reply->type == VALKEY_ERROR_REPLY &&
            reply->len >= prefix_len &&
            memcmp(reply->str, str, prefix_len) == 0);
}
```

Then use it as

```suggestion
    if (replyErrorStartsWith(reply, "MOVED"))
```

zuiderkwast (approved):
Good catch! Good test case too.

src/cluster.c
bjosv: It's a good idea to get rid of the length, but I think its simple enough as it is

PR #306, opened 2026-04-20
Fix issues in valkeyClusterAsyncCallback

bjosv:
Fix several issues in the async cluster command callback:

- NULL dereference on MOVED redirect.
  If `valkeyClusterGetValkeyAsyncContext` failed after a MOVED redirect, `ac_retry` was used without a NULL check, causing a crash.
- Silent callback drop.
  When a retry send or ASKING command failed, the user callback was never invoked, leaving the caller unaware of the failure and unable to clean up `privdata`.
- Removed unnecessary NULL check on `&acc->cc`.

Also restructured the function to use a single exit path, eliminating the `retry:` and `error:` goto labels and making the control flow easier to follow. Added asserts to document invariants and a function-level doc comment.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -3009,9 +3009,12 @@ static void valkeyClusterAsyncCallback(valkeyAsyncContext *ac, void *r,
             if (slot >= 0) {
                 cc->table[slot] = node;
             }
-            ac_retry = valkeyClusterGetValkeyAsyncContext(acc, node);
 
+            ac_retry = valkeyClusterGetValkeyAsyncContext(acc, node);
+            if (ac_retry == NULL)
+                goto done;
             break;
+
         case CLUSTER_ERR_ASK:
             node = getNodeFromRedirectReply(cc, &ac->c, reply, NULL);
             if (node == NULL) {
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -2937,6 +2937,11 @@ static void throttledUpdateSlotMapAsync(valkeyClusterAsyncContext *acc,
     }
 }
 
+/* Callback for async cluster commands. Handles redirects (MOVED/ASK),
+ * retries (TRYAGAIN/CLUSTERDOWN), and delivers the final reply to the user.
+ * On success or unrecoverable failure, the user callback is always invoked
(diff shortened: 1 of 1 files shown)


src/cluster.c
```diff
@@ -2948,27 +2953,26 @@ static void valkeyClusterAsyncCallback(valkeyAsyncContext *ac, void *r,
     valkeyClusterNode *node;
     struct cmd *command;
 
-    if (cad == NULL) {
-        goto error;
-    }
+    assert(cad != NULL);
+    if (cad == NULL)
+        return;
```
zuiderkwast, line 2958: assert + if...? Pick one.

src/cluster.c
```diff
@@ -2948,27 +2953,26 @@ static void valkeyClusterAsyncCallback(valkeyAsyncContext *ac, void *r,
...
 
     acc = cad->acc;
+    assert(acc != NULL);
     if (acc == NULL) {
```
zuiderkwast, line 2962: Same here; either assert or if, not both.

src/cluster.c
```diff
@@ -2948,27 +2953,26 @@ static void valkeyClusterAsyncCallback(valkeyAsyncContext *ac, void *r,
...
-        goto error;
-    }
 
     command = cad->command;
-    if (command == NULL) {
-        goto error;
-    }
+    assert(command != NULL);
+    if (command == NULL)
+        goto done;
```
zuiderkwast, line 2972: And here.

zuiderkwast (commented):
I haven't read the logic carefully yet but I found some mistakes.

src/cluster.c
bjosv: Its due to NDEBUG, get asserts in debug builds and tests, but don't crash in release builds.
Maybe its not worth it? I believe we have some legacy code like it

zuiderkwast: Ah, make sense then! We can keep it.

zuiderkwast: It's a little odd though. You get different behaviour with/without NDEBUG.

zuiderkwast (approved):

PR #312, opened 2026-05-15
Fix unused variable warnings in rdma.c with NDEBUG

### LINKED ISSUE #311: src/rdma.c: unused variable warnings with -DNDEBUG (treated as error with -Werror)
## Description

When building libvalkey v0.5.0 with `-DNDEBUG -Werror`, compilation fails in `src/rdma.c`:

```
src/rdma.c:1039:36: error: variable 'event' set but not used [-Werror=unused-but-set-variable]
 1039 |     struct epoll_event events[1], *event;
      |                                    ^~~~~
src/rdma.c:1038:18: error: unused variable 'ctx' [-Werror=unused-variable]
 1038 |     RdmaContext *ctx = c->privctx;
      |                  ^~~
```

## Root Cause

In the `valkeyRdmaWaitConn` function, both `ctx` and `event` are only used inside an `assert()` call:

```c
event = &events[0];
assert(event->data.fd == ctx->cm_channel->fd);
```

When building with `-DNDEBUG`, the `assert()` macro is removed by the preprocessor, leaving both variables unused. Since the build uses `-Werror`, these warnings become fatal errors.

## Suggested Fix

Add `VALKEY_UNUSED` (already defined in `include/valkey/async.h` and used elsewhere in the same file) to both variable declarations:

```diff
-    RdmaContext *ctx = c->privctx;
-    struct epoll_event events[1], *event;
+    VALKEY_UNUSED RdmaContext *ctx = c->privctx;
+    struct epoll_event events[1];
+    VALKEY_UNUSED struct epoll_event *event;
```

## Reproducer

```bash
cmake . -B build -DCMAKE_BUILD_TYPE=Release -DENABLE_RDMA=1 -G Ninja
ninja -C build
```

NDEBUG is implicitly defined in CMake Release builds.

This issue was found while packaging for conda-forge: https://github.com/conda-forge/libvalkey-feedstock/pull/15

bjosv:
Remove unnecessary local variables `ctx` and `event` in `valkeyRdmaWaitConn` that only existed to make an `assert()` more readable. When built with `-DNDEBUG -Werror`, the assert is removed and both variables become unused, causing a build failure.

Also enable Release build types in the RDMA CI job so that NDEBUG is defined and this class of issue is caught going forward.

Fixes #311

### DIFF
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -1035,8 +1035,7 @@ static int valkeyRdmaCM(valkeyContext *c, long timeout) {
 
 static int valkeyRdmaWaitConn(valkeyContext *c, long timeout) {
     long now, end;
-    RdmaContext *ctx = c->privctx;
-    struct epoll_event events[1], *event;
+    struct epoll_event events[1];
     int nevent;
 
     assert(timeout >= 0);
@@ -1062,8 +1061,9 @@ static int valkeyRdmaWaitConn(valkeyContext *c, long timeout) {
             return VALKEY_ERR;
         }
 
-        event = &events[0];
-        assert(event->data.fd == ctx->cm_channel->fd); /* CM channel fd wakes up only now */
+        /* Only the CM channel fd is polled here */
+        assert(events[0].data.fd == ((RdmaContext *)c->privctx)->cm_channel->fd);
+
         if (valkeyRdmaCM(c, end - now) == VALKEY_ERR) {
             return VALKEY_ERR;
         }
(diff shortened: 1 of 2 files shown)


bjosv:
@pizhenwei Are you ok with this change?

src/rdma.c
```diff
@@ -1035,8 +1035,7 @@ static int valkeyRdmaCM(valkeyContext *c, long timeout) {
 
 static int valkeyRdmaWaitConn(valkeyContext *c, long timeout) {
     long now, end;
-    RdmaContext *ctx = c->privctx;
```
pizhenwei, line 1038: Declaring a temporary variable in a separate line, casting it from a private type pointer, and then accessing it — to me, the original style is more readable. However, I don’t object to the current style either.

Otherwise, LGTM.

pizhenwei (approved):

PR #317, opened 2026-06-02
Fix TLS example argument validation

fallintoplace:
## Summary
- fix the required argument count in `blocking-tls.c`
- fix the required argument count in `async-libevent-tls.c`
- align the optional CA argument checks with the documented usage

## Why
Both examples validated too few required arguments and could read past the validated `argv` range when the certificate or key arguments were missing.

## Validation
- `make -C examples USE_TLS=1 TLS_LDFLAGS='-L/opt/homebrew/opt/openssl/lib -lssl -lcrypto' example-blocking-tls`
- `make -C examples USE_TLS=1 CFLAGS='-fPIC -g -O2 -Wall -Wextra -Werror -I ../include -I/opt/homebrew/opt/libevent/include -L/opt/homebrew/opt/libevent/lib' TLS_LDFLAGS='-L/opt/homebrew/opt/openssl/lib -lssl -lcrypto' example-async-libevent-tls`

### DIFF
--- a/examples/async-libevent-tls.c
+++ b/examples/async-libevent-tls.c
@@ -41,7 +41,7 @@ int main(int argc, char **argv) {
 #endif
 
     struct event_base *base = event_base_new();
-    if (argc < 5) {
+    if (argc < 6) {
         fprintf(stderr,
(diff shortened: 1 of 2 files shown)


bjosv:
Thanks!

PR #320, opened 2026-06-11
Add REUSE 3.3 compliance for machine-readable licensing and SBOM generation

bjosv:
Add `REUSE.md` with per-file copyright and license information,
`LICENSES/` directory with BSD-3-Clause, BSD-2-Clause, and FSFAP license texts,
and a CI workflow to verify compliance.


REUSE.toml
```diff
@@ -0,0 +1,49 @@
...
+path = "src/ffc.h"
+precedence = "aggregate"
+SPDX-FileCopyrightText = "2021, The fast_float authors"
+SPDX-License-Identifier = "MIT"
+
+[[annotations]]
+path = "tests/scripts/simulated-valkey.pl"
+precedence = "aggregate"
+SPDX-FileCopyrightText = "2020, Ericsson Software Technology <viktor.soderqvist@est.tech>"
+SPDX-License-Identifier = "FSFAP"
```
zuiderkwast, line 49: Is this the only file that's using the FSFAP license? I can relisence it to MIT or BSD so we can get rid of FSFAP...

zuiderkwast (approved):
Should we include the copyright lines for src/crc16.c in the reuse file? It has a copyright holder "2001-2010 Georges Menie" and Salvatore. Or is that covered by the `precedence = "aggregate"` rule?

REUSE.toml
zuiderkwast: Or we can keep it. The license is extremely small.

bjosv: If we keep the FSFAP license then the linter in the valkey repo requires us to add FSFAP.txt to that repo as well when vendoring.
I think it would be good to reduce the number of licenses. WDYT?

zuiderkwast: OK, I'll do it next week.

PR #323, opened 2026-06-22
Add c-ares support for DNS resolution with timeout

bjosv:
`getaddrinfo()` has no timeout mechanism. When DNS is slow or unresponsive, libvalkey hangs indefinitely with no way to cancel the lookup.

This PR adds optional c-ares integration (`USE_CARES=1` / `-DENABLE_CARES=ON`) that replaces `getaddrinfo()` with c-ares driven by a `poll()` loop. DNS resolution respects `connect_timeout` (defaulting to 5 seconds if unset), so lookups that previously hung indefinitely now fail with a timeout error.

The PR consists of two commits: the first moves existing DNS logic into `dns.[ch]`, and the second adds the c-ares integration.

**Notes:**
- The async API currently still blocks during DNS (now with a timeout). A follow-up PR
  #324
will make DNS fully non-blocking for the async API via event-loop integration.
- Without c-ares enabled, behavior is unchanged (plain `getaddrinfo`)
- IPv4/IPv6 preference flags and fallback behavior are preserved
- IPv6 literals (e.g. `::1`) are detected and handled correctly
- Targets Linux, macOS, FreeBSD (not Windows)

**Testing:**
- New `ut_dns_cares` unit test covering IP literals, hostnames, non-existent hosts, IPv6, and timeout
- OOM handling test updated to accommodate additional allocations from c-ares
- `ut_connect_fallback` disabled when c-ares is enabled (it overrides `getaddrinfo` which c-ares bypasses)
- New CI job with sanitizers and IPv6 tests enabled


src/dns.c
```diff
@@ -0,0 +1,342 @@
...
+            break;
+        }
+
+        struct timeval maxtv, tv;
+        maxtv.tv_sec = remaining / 1000;
+        maxtv.tv_usec = (remaining % 1000) * 1000;
+        struct timeval *tvp = ares_timeout(channel, &maxtv, &tv);
+        int poll_ms = (int)(tvp->tv_sec * 1000 + tvp->tv_usec / 1000);
+        if (poll_ms <= 0)
+            poll_ms = 1;
```
michael-grunder: Not sure if we need to be this paranoid though

```suggestion
        long lval_ms = tvp->tv_sec * 1000 + tvp->tv_usec / 1000;
        if (lval_ms <= 0)
              lval_ms = 1;
        else if (lval_ms > INT_MAX)
            lval_ms = INT_MAX;
        int poll_ms = lval_ms;
```

src/dns.c
```diff
@@ -0,0 +1,342 @@
...
+static int valkeyResolveCares(const char *host, int port, int flags,
+                              long timeout_ms, struct addrinfo **result) {
+    ares_channel_t *channel = NULL;
+    struct ares_options opts;
+    struct ares_addrinfo_hints hints;
+    struct caresResult res = {0, 0, NULL};
+    struct caresSockState sockstate = {{{0}}, 0};
+    int optmask;
+    int rv;
+    long effective_timeout = (timeout_ms > 0) ? timeout_ms : VALKEY_DNS_DEFAULT_TIMEOUT_MS;
```
michael-grunder: This depends on what we want `valkeyConnect` to do. I think it sets the timeout to `INT_MAX` which would not cause the previous code to set the 5s default.

```suggestion
    long effective_timeout = timeout_ms;
    if (effective_timeout <= 0 || effective_timeout >= INT_MAX)
        effective_timeout = VALKEY_DNS_DEFAULT_TIMEOUT_MS;

```

michael-grunder (approved):
PR looks awesome.

src/dns.c
bjosv: Ooh, good call. The timeout in c-ares became ~24 days instead of 5s default.
Yeah, I went with 5s that should be generous enough for DNS, I believe..

PR #327, opened 2026-06-23
Replace strncpy with length-validated memcpy for Unix socket path

bjosv:
Previously, paths longer than sun_path (108 bytes on Linux) were silently truncated via strncpy, potentially connecting to the wrong socket.

Now returns an error if the path doesn't fit. Also switched to vk_calloc to ensure the sockaddr_un is zero-initialized.


michael-grunder (approved):
Interesting. I didn't know this was a limit.

bjosv:
> Interesting. I didn't know this was a limit.

First I only planned to fix strncpy from what was mentioned in #325 , but I also learnt about it while digging

PR #328, opened 2026-06-26
Fix async command timeout never firing under continuous writes

bjosv:
The command timeout timer was reset on every write, so continuously sending commands to an unresponsive server would prevent the timeout from ever firing.
    
Fixed by not resetting the timer if its is already active. When the timer fires, check if any replies (including push messages) were received since it started. If yes, the server is alive and the timer restarts. If no, it's a real timeout and the connection is closed.

Additional issue in the `libevent` adapter (used in test):
The timeout was combined with the I/O event in a single struct event. With the needed `EV_PERSIST`, libevent resets the timeout on every I/O callback. Even if the library didn't reschedule, any read/write activity silently restarted the timer.

### DIFF
--- a/include/valkey/adapters/libevent.h
+++ b/include/valkey/adapters/libevent.h
@@ -42,6 +42,7 @@
 typedef struct valkeyLibeventEvents {
     valkeyAsyncContext *context;
     struct event *ev;
+    struct event *timer;
     struct event_base *base;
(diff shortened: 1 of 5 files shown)


michael-grunder (approved):
LGTM

This edge case has been around forever too.

PR #330, opened 2026-07-03
Add internal timer system for async timeout handling

bjosv:
Adds a pre-allocated timer pool that supports multiple concurrent async timers through a single adapter hook.
This allows adding additional timers (cluster topology refresh, c-ares DNS timeouts, health checks) without requiring adapter changes.

**Timer framework (commit 1)**
- Pre-allocated pool of timers, sorted linked list by deadline
- One timer processed per call (safe if callback frees the context)
- Uses CLOCK_MONOTONIC / QueryPerformanceCounter

**Async timeout migration (commit 2)**
- Separate `connect_timer` and `command_timer` with own callbacks
- Connect timer cancelled on successful connect
- Command timer tracks reply count to detect unresponsive servers
- `refreshTimeout` is now a regular exported function (for TLS .so)
- No adapter changes

**Behavior change**
Changing `command_timeout` at runtime no longer affects an already-running timer.
The new value takes effect after the current timer fires.

### DIFF
--- a/src/async.c
+++ b/src/async.c
@@ -46,6 +46,7 @@
 #include "async_private.h"
 #include "dict.h"
 #include "net.h"
+#include "timer.h"
 #include "valkey_private.h"
 #include "vkutil.h"
 
@@ -152,7 +153,10 @@ static valkeyAsyncContext *valkeyAsyncInitialize(valkeyContext *c) {
     ac->sub.schannels = schannels;
     ac->sub.pending_unsubs = 0;
 
-    ac->timeout_reply_count = VALKEY_TIMEOUT_INACTIVE;
+    ac->timer_list = NULL;
+    ac->connect_timer = NULL;
+    ac->command_timer = NULL;
+    ac->timeout_reply_count = 0;
 
     return ac;
 oom:
@@ -377,6 +381,13 @@ static void valkeyAsyncFreeInternal(valkeyAsyncContext *ac) {
         dictRelease(ac->sub.schannels);
     }
 
+    /* Free internal timers. */
+    if (ac->timer_list) {
+        valkeyTimerListFree(ac->timer_list);
+        vk_free(ac->timer_list);
+        ac->timer_list = NULL;
+    }
+
     /* Signal event lib to clean up */
     _EL_CLEANUP(ac);
 
@@ -586,7 +597,7 @@ void valkeyProcessCallbacks(valkeyAsyncContext *ac) {
             c->flags |= VALKEY_SUPPORTS_PUSH;
 
         /* Any data from the server means it's alive. */
-        if (ac->timeout_reply_count != VALKEY_TIMEOUT_INACTIVE)
+        if (ac->command_timer != NULL)
             ac->timeout_reply_count++;
 
         /* Send any non-subscribe related PUSH messages to our PUSH handler
@@ -695,6 +706,10 @@ static int valkeyAsyncHandleConnect(valkeyAsyncContext *ac) {
          * to disconnect.  For that reason, permit the function
          * to delete the context here after callback return.
          */
+        if (ac->connect_timer) {
+            valkeyTimerDel(ac->timer_list, ac->connect_timer);
+            ac->connect_timer = NULL;
+        }
         c->flags |= VALKEY_CONNECTED;
         valkeyRunConnectCallback(ac, VALKEY_OK);
         if ((ac->c.flags & VALKEY_DISCONNECTING)) {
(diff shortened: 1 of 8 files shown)


src/timer.c
```diff
@@ -0,0 +1,162 @@
...
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    tv->tv_sec = ts.tv_sec;
+    tv->tv_usec = (int)(ts.tv_nsec / 1000);
+#else
+    LARGE_INTEGER counter, frequency;
+    QueryPerformanceCounter(&counter);
+    QueryPerformanceFrequency(&frequency);
+    int64_t usec = counter.QuadPart * 1000000 / frequency.QuadPart;
+    tv->tv_sec = (long)(usec / 1000000);
+    tv->tv_usec = (long)(usec % 1000000);
```
michael-grunder:
```suggestion
    int64_t sec = counter.QuadPart / frequency.QuadPart;
    int64_t rem = counter.QuadPart % frequency.QuadPart;
    tv->tv_sec = (long)sec;
    tv->tv_usec = (long)(rem * 1000000 / frequency.QuadPart);
```

michael-grunder:
This is a massive improvement.

I'm still going through it but I have a couple extra commits [in a branch on my fork](https://github.com/valkey-io/libvalkey/compare/main...michael-grunder:libvalkey:mine/add-timer-system?expand=1)

Just the overflow comment and re-allowing a user to disable a timeout after it has been set. I think we still want that behavior?

src/async.c
```diff
@@ -777,54 +792,126 @@ void valkeyAsyncHandleWrite(valkeyAsyncContext *ac) {
...
+    }
+}
+
+/* Called by adapters when the scheduled timer expires. Dispatches internal
+ * timers and reschedules the adapter if more timers are pending. */
+void valkeyAsyncHandleTimeout(valkeyAsyncContext *ac) {
+    valkeyContext *c = &(ac->c);
+    struct timeval remaining;
+    /* must not be called from a callback */
+    assert(!(c->flags & VALKEY_IN_CALLBACK));
```
michael-grunder, line 909:
```suggestion
    assert(!(c->flags & VALKEY_IN_CALLBACK));
    (void)c;
```

src/async.c
```diff
@@ -777,54 +792,126 @@ void valkeyAsyncHandleWrite(valkeyAsyncContext *ac) {
...
 }
 
-void valkeyAsyncHandleTimeout(valkeyAsyncContext *ac) {
+/* Add a timer and notify the adapter if rescheduling is needed. */
+valkeyTimer *valkeyAsyncAddTimer(valkeyAsyncContext *ac, struct timeval timeout,
+                                 valkeyTimerProc proc, void *privdata) {
+    if (ac->timer_list == NULL) {
+        ac->timer_list = vk_malloc(sizeof(valkeyTimerList));
+        if (ac->timer_list == NULL)
+            return NULL;
```
michael-grunder: Should we allocate this eagerly in `valkeyAsyncInitialize`? It's ~160 bytes or so but that makes the error path a lot nicer. Right now if this allocation fails no timer is set and a client could hang indefinitely

src/timer.c
bjosv: oh, overflow, thanks

src/async.c
bjosv: hum, yes you are right, I will look at it.

bjosv:
@michael-grunder Moved the valkeyTimerList allocation now, alot better. I also added your improvement commits.

I wonder if its possible to get rid of `refreshTimeout` and its call in every `_EL_ADD_READ`/`_EL_ADD_WRITE` somehow now.
Previously it handled the multiplexing of timers, but now we might just be able to start each timer when connecting and when sending a command.

I think the command timeout could be moved already by starting the `command_timer` in `valkeyAsyncAppendCmdLen` instead.

The connect timer is harder since we need to start it in `valkeyAsyncConnect*`, **or**  when the event-adapter is attached by the user via calls like `valkeyPollAttach()`. Maybe this was the reason for doing it in `refreshTimeout` from the start?

michael-grunder (approved):
LGTM. I'm sorry about the delay :sob:

bjosv:
@michael-grunder Thanks, now worries. When taking another look I found an issue which is now fixed in the last commit (needed later for async DNS). PTAL
The framework supports multiple concurrent timers, but the adapter wasn't re-armed when a timer was fired and there were more timers pending. Added some tests as well.

PR #331, opened 2026-07-03
Add option to attach an adapter in valkeyAsyncConnectWithOptions()

bjosv:
Add `attach_fn` and `attach_data` to `valkeyOptions`, allowing the event-loop adapter to be specified as part of the connection options. When set, `valkeyAsyncConnectWithOptions()` attaches the adapter internally after the connect is initiated.

This mirrors the existing cluster API pattern (`valkeyClusterOptions`) and enables future work where the adapter drives non-blocking DNS resolution.

The legacy path (attaching the adapter separately after connect) remains fully supported.

**Changes**
- Adds `attach_fn` / `attach_data` fields in `valkeyOptions`
- For TCP: connect is deferred (`VALKEY_CONNECT_DEFERRED`) until after context setup, then DNS + connect runs before adapter attach
- For Unix: connect completes immediately, adapter attached directly


michael-grunder (approved):
LGTM

Edit: Deleted comment with `s/above/below/` I misread the sentence.

bjosv:
Is there any better names than `attach_fn`/`attach_data` maybe?
We use these names in `valkeyClusterOptions` too, but not sure if its descriptive enough..

I also need to create another PR to be able to set connect callbacks, something like
```
   /* Optional async connect/disconnect callbacks. When set,
     * valkeyAsyncConnectWithOptions registers them on the context before the
     * connect is initiated, so no connect event can be missed even when the
     * event loop is already running. Equivalent to calling
     * valkeyAsyncSetConnectCallback / valkeyAsyncSetDisconnectCallback. */
    void (*async_connect_callback)(struct valkeyAsyncContext *ac, int status);
    void (*async_disconnect_callback)(const struct valkeyAsyncContext *ac, int status);
```

michael-grunder:
> Is there any better names than attach_fn/attach_data maybe?

Seems like reasonable names to m,e but I'm not great at naming things 😄

bjosv:
> Seems like reasonable names to m,e but I'm not great at naming things 😄

Me either :) I'll merge this and prepare a follow-up PR regarding options for setting callbacks and tls_init_fn

PR #333, opened 2026-07-07
cluster: initialize primary_id to NULL in parse_cluster_nodes

benh57:
We noticed this because we forked libvalkey locally and built in our own env

primary_id is declared uninitialized and only written by parse_cluster_nodes_line() when replicas are parsed (a non-NULL parsed_primary_id is passed, i.e. add_replicas).

It is later read by retain_replica_node() on replica lines. A replica line is only produced when add_replicas is set, so in practice the value is written before it is read, but that correlation is not visible to the compiler, which warns about a possibly-uninitialized read (-Wmaybe-uninitialized / MSVC /W4, fatal under -Werror). 

Initialize to NULL to silence the warning and guard against future refactors.


michael-grunder (approved):

bjosv:
Thanks! Should we enable it in CI , or is there a risk of false positives?

PR #336, opened 2026-07-22
Add `maxdepth` member to `valkeyReader`.

michael-grunder:
Currently malicious `RESP` protocol data consisting of an extremely nested reply can cause a stack overflow in the `valkeyReader` parsing logic.

This commit simply adds a new member `maxdepth` which is very similar to the existing `maxelements` member. By default it limits the maximum nesting of `RESP` replies to 1024 which should be much higher than normal payloads reach while being far below the stack overflow risk.


bjosv (approved):
Nice! Limiting to 1024 nested array/map/set/attr/push should be fine

PR #343, opened 2026-09-04
Fix potential signed overflow plus better Windows function

michael-grunder:
* Fixed potential signed overflow in `valkeyPollMillis()` and `valkeyDnsPollMillis()`
* Use `GetTicksCount64` on windows since it directly gives us elapsed milliseconds since boot.
* Fix an off-by-one calculation when guarding against invalid timeouts

### DIFF
--- a/src/sockcompat.c
+++ b/src/sockcompat.c
@@ -258,7 +258,7 @@ int win32_setsockopt(SOCKET sockfd, int level, int optname, const void *optval, 
         const struct timeval *tv = optval;
         DWORD timeout;
         uint64_t timeout_msec;
-        if (tv->tv_sec < 0 || tv->tv_usec < 0 || tv->tv_usec > 1000000) {
+        if (tv->tv_sec < 0 || tv->tv_usec < 0 || tv->tv_usec >= 1000000) {
             errno = EINVAL;
             return -1;
         }
--- a/src/sockcompat.c
+++ b/src/sockcompat.c
@@ -259,11 +259,13 @@ int win32_setsockopt(SOCKET sockfd, int level, int optname, const void *optval, 
         DWORD timeout;
         uint64_t timeout_msec;
         if (tv->tv_sec < 0 || tv->tv_usec < 0 || tv->tv_usec >= 1000000) {
+            WSASetLastError(WSAEINVAL);
             errno = EINVAL;
             return -1;
         }
         timeout_msec = (uint64_t)tv->tv_sec * 1000 + (uint64_t)tv->tv_usec / 1000;
         if (timeout_msec > MAXDWORD) {
+            WSASetLastError(WSAEINVAL);
             errno = EINVAL;
             return -1;
         }
--- a/src/sockcompat.c
+++ b/src/sockcompat.c
@@ -32,6 +32,8 @@
 #include "sockcompat.h"
 
 #ifdef _WIN32
+#include <stdint.h>
+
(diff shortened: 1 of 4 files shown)


bjosv (approved):
Nice.
When grepping for timeval, is there a similar issue for the windows DWORD in
https://github.com/valkey-io/libvalkey/blob/5c6de8845511c8c152925fd2208bf0d0a12f3d1f/src/sockcompat.c#L257

and then there is the order of multiply and remainder that could be switched
https://github.com/valkey-io/libvalkey/blob/5c6de8845511c8c152925fd2208bf0d0a12f3d1f/src/sockcompat.c#L233

michael-grunder:
I fixed the overflow risk in sockcompat.c as well.

Note that the Windows changes were mostly written by a clanker :smile:. All I did was clean up extra typecasts  they like to do.

src/sockcompat.c
```diff
@@ -254,8 +256,19 @@ int win32_setsockopt(SOCKET sockfd, int level, int optname, const void *optval,
     int ret = 0;
     if ((level == SOL_SOCKET) && ((optname == SO_RCVTIMEO) || (optname == SO_SNDTIMEO))) {
         const struct timeval *tv = optval;
-        DWORD timeout = tv->tv_sec * 1000 + tv->tv_usec / 1000;
-        ret = setsockopt(sockfd, level, optname, (const char *)&timeout, sizeof(DWORD));
+        DWORD timeout;
+        uint64_t timeout_msec;
+        if (tv->tv_sec < 0 || tv->tv_usec < 0 || tv->tv_usec > 1000000) {
```
bjosv: To match the other linux change?
```suggestion
        if (tv->tv_sec < 0 || tv->tv_usec < 0 || tv->tv_usec >= 1000000) {
```

src/sockcompat.c
```diff
@@ -254,8 +256,19 @@ int win32_setsockopt(SOCKET sockfd, int level, int optname, const void *optval,
...
+            errno = EINVAL;
+            return -1;
+        }
+        timeout_msec = (uint64_t)tv->tv_sec * 1000 + (uint64_t)tv->tv_usec / 1000;
+        if (timeout_msec > MAXDWORD) {
+            errno = EINVAL;
```
bjosv, line 269: An AI, that probably "knows" windows better than me, told me that we should use 
```
WSASetLastError(WSAEINVAL);
_updateErrno(0);
return -1;
```
so that a user who uses `WSAGetLastError()` gets this error and not the previous one.
Same above.

michael-grunder: Believe it or not the AI did this as well the first revision, but I opted for a smaller structural change since I know nothing about windows. I'll go ahead and update it to include that logic. People will tell us if it's an issue.

michael-grunder: > OK so we don't want _updateErrno(0)?  That's fine I just want to have a bit of understanding
> • Right—we don’t want _updateErrno(0) here.
> That helper means “the socket operation failed; read WSAGetLastError() and translate it into errno.” But its translation table
  doesn’t handle WSAEINVAL, so it falls back to EIO, overwriting our correct EINVAL.

It's a brave new world :smile:

bjosv: Oh, it falls back to EIO.. 
Looks great. I'll merge

bjosv (approved):

PR #345, opened 2026-09-15
fix: Harden aggregate allocation

michael-grunder:
Currently libvalkey eagerly allocates aggregates (array or map types) up front, so when we receive the resp `*4294967295\r\n` string will attempt to allocate an enormous array of elements.

This commit does two things:

* Reduces the default `maxelements` to 2^26 (67 million elements)
* Switches to a maximum initial allocation of 1024 elements and then doubles each time we hit the capacity, until doubling would be too large at which point we allocate exactly total.


michael-grunder:
This is another parsing hardening PR that applies to libvalkey.

Right now `*4294967295\r\n` will cause libvalkey to allocate all of those array elements upfront meaning a tiny RESP header could be used to crash or otherwise stall any clients.

Another much smaller surface area change might be to just reduce the maxelements default.

A clanker wanted to add a huge amount of logic to `client_test.c` but it seemed excessive. If you want I can commit those tests as well.

bjosv (approved):
Nice. Just the attachReplyObject was a nice cleanup.
The asserts could be a soft spot since they are noop in release builds, but its same as legacy.

The failing CI should be fixed by #348

PR #346, opened 2026-09-15
fix: Protect against overflow on LLP64

michael-grunder:
Protect against overflow on LLP64.

Again there is an extensive test the clanker wanted to add but it seemed like too much noise.


michael-grunder:
The test failure seems unrelated to the PR. It's a package problem in the runner.

bjosv (approved):
The pesky LLP64..

PR #350, opened 2026-09-16
fix: Harden async subscribe reply shape logic

michael-grunder:


michael-grunder:
I did this one in two commits because the second commit changes from using an assertion which can be elided with -DNDEBUG to an error reply.

It seems like nicer ergonomics but it does change behavior.

src/async.c
```diff
@@ -498,7 +498,7 @@ static int valkeyGetSubscribeCallback(valkeyAsyncContext *ac, valkeyReply *reply
      * The type and number of elements (3 to 4) are specified at:
      * https://valkey.io/docs/topics/pubsub/#format-of-pushed-messages */
     if ((reply->type == VALKEY_REPLY_ARRAY && !(c->flags & VALKEY_SUPPORTS_PUSH) && reply->elements >= 3) ||
-        reply->type == VALKEY_REPLY_PUSH) {
+        (reply->type == VALKEY_REPLY_PUSH && reply->elements >= 3)) {
         assert(reply->element[0]->type == VALKEY_REPLY_STRING);
```
bjosv: We should probably convert this assert to a real guard for release builds, avoiding a NULL deref if its an integer. This could be part of the match; requiring the first element to be a string.

src/async.c
bjosv: This wont crash if it's a non-integer I think, maybe we should handle it like the unknown_callback case?

bjosv (approved):
Nice, the pubsub is a bit messy from start, but hardening is great.
I found 2 additional items while looking, maybe can be done in this PR, but you decide.

michael-grunder:
Moving away from assertions should make things nicer for users.

bjosv (approved):
Very nice

PR #352, opened 2026-09-18
fix: Harden RDMA recv against peer-supplied length

bjosv:
The length passed to `connRdmaHandleRecvImm()` is the immediate value chosen by the peer, but it was only checked with an assertion, which is elided with `-DNDEBUG`.
Check it against the space remaining in the receive buffer and return an error instead, comparing against the space left rather than summing since `byte_len + rx_offset` can wrap.

Found during an assert inventory to identify asserts that can be externally triggered.


bjosv:
@pizhenwei Does this seem right?

pizhenwei (approved):
LGTM.

`RdmaContext *ctx = c->privctx;`, so only `valkeyContext *c` is also fine for `connRdmaHandleRecvImm`. Both styles are fine to me.

bjosv:
> LGTM.
> 
> `RdmaContext *ctx = c->privctx;`, so only `valkeyContext *c` is also fine for `connRdmaHandleRecvImm`. Both styles are fine to me.

Great, thanks! I kept the style so its the same as in `connRdmaHandleRecv`.

